PR #1, opened 2022-06-16
Speed up prediction

jasonrute:
Speed up the prediction function.


pestun:
Thank you, will test and profile

jasonrute:
We don't need this.

PR #2, opened 2022-06-16
updated document line for Predict()

pestun:
the values returned by Predict() are predicted probabilies.

The sum of values is <= 1.0 since not all possible options are returned.


jasonrute (approved):
LGTM!  Thanks for cleaning this up!

PR #4, opened 2022-06-17
bfs_option default is now True

pestun:
seems semantically more meaningful to cut out large graphs by depth level since graph neural network operates at finite depth


fidel-schaposnik (approved):
Looks good, nothing breaks on TF-GNN side since the output format stays the same

pestun:
i'll merge

PR #5, opened 2022-06-17
Vasily/loader refactor

pestun:
- continue to remove the legacy loader methods 

- removed uint64 packed representation of global node (file_idx, node_idx):
    - from now on, python uses either python tuple(int,int) to represent a global node (if we want representation to be python-hashable to put into a dict) or np.array of shape=(2,), dtype=np.uint32 if we want to send the node to low-level c++ method
    - c++ uses c_GlobaNode {uint32, uint32} structure for a global node through the code with the support for hash and comparison methods to be used in STL hash maps and ordered maps


pestun:
( to reduce the number of different representations and internal data structures for readability and maintenance of the code)

fidel-schaposnik (approved):
Looks good!

PR #7, opened 2022-06-18
Vasily/tfgnn support

pestun:


fidel-schaposnik (approved):
Looks good!

PS: Root is currently not used in the GNN, so it is not encoded in the GraphTensor for proof-states... this should be changed if we ever decide to use it somehow

LasseBlaauwbroek:
@fidel-schaposnik How does your model currently determine the final embedding of a graph? Using some kind of average of all embeddings of the individual nodes in the graph? (I ask because you say that you're not using the root.)

fidel-schaposnik:
@LasseBlaauwbroek The final embedding of the graph is computed by a pooling layer from the embeddings from all its nodes. The type of pooling is a hyperparameter (max seems to work best, but we've also tried mean and attention pooling), but the root node is not singled out at this stage (which does not mean the GNN itself cannot detect and single it out, if it wants to). I could think of various ways in which to include the root node information in the pooling process, but haven't really explored any of them so I wouldn't know how they compare to the current approach...

pestun:
Merging!

PR #10, opened 2022-06-23
Vasily/hmodel

pestun:
added oracle graph-hash based model

this model hashes deterministically the graphs based on the ordering of children (outlinks) from each node, and records rigidly the set of possible actions for each proofstate based on a hash of proofstate. 

this  model models perfectly overtrained network with no-definition-update table and alignment by global definition names


pestun:
Jason, shall we merge this ?

PR #22, opened 2022-07-08
Consolidation of the Predict API

fidel-schaposnik:
@jasonrute Could you please check that this doesn't break evaluation on the TF2 model, and if so merge into main?

### DIFF
--- a/graph2tac/predict.py
+++ b/graph2tac/predict.py
@@ -1,0 +1,115 @@
+from typing import List, Tuple, Optional
+
+import numpy as np
+
+from graph2tac.loader.data_server import GraphConstants
+
+
+NUMPY_NDIM_LIMIT = 32
+
+
+def cartesian_product(*arrays):
+    """
+    using the code from  https://stackoverflow.com/questions/11144513/cartesian-product-of-x-and-y-array-points-into-single-array-of-2d-points
+    """
+    la = len(arrays)
+
+    if la > 32:
+        print(arrays)
+    dtype = np.result_type(*arrays)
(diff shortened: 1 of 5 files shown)


graph2tac/predict.py
```diff
@@ -0,0 +1,115 @@
...
+class Predict:
+    """
+    Common prediction API to load training checkpoints and make predictions in order to interact with an evaluator.
+    This class exposes the following methods:
+        - `initialize`: set the global context during evaluation
+        - `compute_new_definitions`: update node label embeddings using definition cluster graphs
+        - `ranked_predictions`: make predictions for a single proof-state
+        - `get_tactic_index_to_numargs`: access the value of `tactic_index_to_numargs` seen during training
+        - `get_tactic_index_to_hash`: access the value of `tactic_index_to_hash` seen during training
+        - `get_node_label_to_name`: access the value of `node_label_to_name` seen during training
```
pestun, line 35: was renamed

pestun:
looks good!

graph2tac/predict.py
fidel-schaposnik: Yes, I renamed the functions `get_node_label_to_name` and `get_node_label_in_spine` to `get_label_to_name` and `get_label_in_spine` to match how they appear in the `GraphConstants` object, see https://github.com/IBM/graph2tac/blob/fa2d2c406e1907547a4e4c0b819a0abc42cc0c2f/graph2tac/loader/data_server.py#L43 . Should we keep these names, or maybe I should rename the attributes of the `GraphConstants` to be `node_label_to_name` and `node_label_in_spine`?

PR #42, opened 2022-08-02
Fidels/normalization fix

fidel-schaposnik:
We have to make sure that the loss computation does not independently normalize logits for global and local arguments!


LasseBlaauwbroek:
Do you think this improves the global argument predictions?

PR #51, opened 2022-08-08
Add tqdm and numpy to setup.py

fidel-schaposnik:
Both packages are explicitly used in our code, yet they were not included in setup.py. I've added them so that IDEs don't complain.

Signed-off-by: Fidel I. Schaposnik Massolo <fidel.s@gmail.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,8 @@ setup(
         'keras>=2.8',
         'tensorflow>=2.8',
         'tensorflow_gnn>=0.2.0.dev1',
+        'tqdm',
+        'numpy',
         'fire',
         'pycapnp',
         'psutil',


jasonrute (approved):
LGTM.  I admit, I don't know why pycharm needs an explicit dependency.  (I guess vs code just looks at whatever is available to figure this out.)  But I'm happy to make these explicit.

fidel-schaposnik:
I guess the issue is in principle you cannot know from setup.py whether they would be installed as dependencies of other packages, so a priori the code is not guaranteed to run (and even if it currently runs, if in the future they got dropped from all dependencies it would stop running without any action on our part)

PR #57, opened 2022-08-16
Mirek/definition strings

mirefek:
Optional definition names + possible to configure


LasseBlaauwbroek:
Great Mirek! What is the next step on this? Are we ready to test this on ministdlib? Or did you already do this?

fidel-schaposnik:
I think the next step is for me to review these changes before merging into main, but in the meantime tests on mini-stdlib can be run if they haven't been run already...

fidel-schaposnik:
As a side note, @mirefek can you check your git sign-off workflow? It looks like GitHub is unhappy with "Olk Miroslav <olsak@ihes.fr>" in place of "Mirek Olsak <mirek@olsak.net>", and some other commits are unsigned (note that merging another branch into the branch you are working on involves creating a commit, so you have to `git commit --amend --signoff` immediately afterwards)

mirefek:
> As a side note, @mirefek can you check your git sign-off workflow? It looks like GitHub is unhappy with "Olk Miroslav [olsak@ihes.fr](mailto:olsak@ihes.fr)" in place of "Mirek Olsak [mirek@olsak.net](mailto:mirek@olsak.net)", and some other commits are unsigned (note that merging another branch into the branch you are working on involves creating a commit, so you have to `git commit --amend --signoff` immediately afterwards)

I thought it is only unhappy with the unsigned commits (I forgot to sign them some time ago, and I didn't properly understood how to fix them). How do you see that it is also unhappy with the Ols Miroslav?

fidel-schaposnik:
Maybe see if this link works for you:

https://github.com/IBM/graph2tac/pull/57/checks?check_run_id=7860851275

The summary at the bottom says:

Commit sha: [a0ce2c0](https://github.com/IBM/graph2tac/pull/57/commits/a0ce2c025b632c466bfc4c1d08480f440c1b8f43), Author: Mirek Olsak, Committer: Mirek Olsak; Expected "Mirek Olsak [mirek@olsak.net](mailto:mirek@olsak.net)", but got "Olk Miroslav [olsak@ihes.fr](mailto:olsak@ihes.fr)".
Commit sha: [5d2a24f](https://github.com/IBM/graph2tac/pull/57/commits/5d2a24fc635ac9936bb45d073032de46ab948e95), Author: Olk Miroslav, Committer: Olk Miroslav; The sign-off is missing.
Commit sha: [945e6d9](https://github.com/IBM/graph2tac/pull/57/commits/945e6d9ea0f898d848fe0ed1fe4384df014584e8), Author: Olk Miroslav, Committer: Olk Miroslav; The sign-off is missing.
Commit sha: [b246950](https://github.com/IBM/graph2tac/pull/57/commits/b24695073b32d5cbeb689e0c8c068ba4b1c89e61), Author: Olk Miroslav, Committer: Olk Miroslav; The sign-off is missing.

(Note that there's also instructions to fix this)

jasonrute:
I found a number of issues @mirefek and @fidel-schaposnik.  The issues aren't so much with the previous code setting up the tokenizer.

Technically if we are using a tokenizer without an explicit vocabulary, that is a **trained model**.   (And we do appear to be training the model [here](https://github.com/IBM/graph2tac/blob/bbe3a76dd0023e74c3bf076ca6f87f915d4c5a7f/graph2tac/tfgnn/dataset.py#L65) with `adapt`.) In particular:

- The tokenizer should be be trained on only the training set.
  - As far as I can see it is being trained on all the data.
- When we save our model we also need to save the tokenizer.
  - I can’t see where it is being saved
- When we load our model we should load the tokenizer (instead of retraining it).
  - I can’t see where it is being loaded.
  - In training it appears to re-trained from the data (if say one is restarting the training from a checkpoint).
  - Retraining the tokenizer and expecting the same outcome requires that both `adapt` is deterministic and the dataset is the same as with the previous checkpoint.  These are not safe assumptions.
- When we apply our model to new data (as we do when updating definitions during benchmarking) we apply the trained tokenizer to the data.
  - The only place that `tokenize_definition_graph` is being called is in training.
  - I think the definition embeddings recreation would crash since the model would expect “definition_name_vectors” in the definition graph, but it wouldn't be there.
  - `tokenize_definition_graph` should also be called everywhere we create a tfgnn definition graph consistantly.  I can’t find the specific code which does that, but it should be bundled with that code instead of something we do much later.

Another more radical approach is not to vectorize the names at the point of the data loader, but to do it inside the model.  So we will be passing strings to the model instead of vectors of integers.  That way the tokenizer would live in the model and be part of the model weights.  When one loads the model, they would also load the tokenizer.  One still needs to train a tokenizer before training the model.  But after that is done the tokenizer itself would be part of the model.  I don’t know if this is silly or not.  (I doubt tokenizing strings would slow down the definition task.)  But if we don't do this, then we would need to save and load the tokenizer alongside the model.

Another quick fix would be to use a fixed vocabulary for the tokenizer, so no training is required.  Then we wouldn’t train it with `adapt`.  Most of my bullets above would be null and void.  The only thing we would still need to do, is tokenize the definition graphs coming from recomputing definition embeddings in the server (and other places we recompute definition embeddings).

graph2tac/tfgnn/models.py
```diff
@@ -803,29 +804,53 @@ def __init__(self,
...
         self._hidden_size = hidden_size
 
+        if name_layer is None:
+            self._name_layer = None
+        else:
+            self._name_layer_core = tf.keras.layers.deserialize(name_layer)
+            self._name_layer = tf.keras.Sequential([
+                tf.keras.layers.Input(shape=[None], dtype=tf.float32, ragged=True),
+                tf.keras.layers.Embedding(
+                    input_dim=Dataset.MAX_LABEL_TOKENS,
```
jasonrute, line 814: I'm worried about you getting this number from here.  The size we should use should be based on whatever that constant was from when we trained the model.  I would put this number in the dataset graph constants and get it from there.  (Or if we put the tokenizer in the model, you could get it from the tokenizer.)

jasonrute (changes requested):
Mostly looks good.  The largest concern is with the tokenizer as I mention in another message.

fidel-schaposnik:
Putting the tokenizer inside the model would indeed be more elegant, and it was @mirefek 's original choice too, but unfortunately you cannot use `tf.RaggedTensor` objects of type `tf.string` in conjunction with `tf.distribute.MirroredStrategy` (_i.e._ with multi-GPU training). This is a problem because our graph clusters have a variable number of definition names, so they would produce such objects when batched if you simply include the definition names in the graph (instead of their tokenization, which results in `tf.RaggedTensor` objects of type `tf.int32`).

The solution I found was to move the tokenizer to the input pipeline, which runs on the CPU anyway, but indeed this brings its own bag of (more minor) issues. I think the simplest solution, at least in order to test whether this idea helps at all, is to use a fixed vocabulary for the tokenizer and then mostly forget about it. Otherwise, one should indeed pay close attention to how the tokenizer is trained, saved and reloaded.

Finally, I was under the impression that @mirefek had already implemented the re-computation of new definitions during evaluation, but indeed it seems strange that no tokenization is going on. I'd say that at least `tokenize_definition_graph` should be called here: https://github.com/IBM/graph2tac/blob/1cc028c46449c53fa724d90b74e85e465291a8c6/graph2tac/tfgnn/predict.py#L308 . Then hopefully if the model was saved with a layer for definition names the rest should be mostly fine, though that remains to be checked...

jasonrute:
Ok, so for now, I think Fidel is saying the simplest approach is:
* To use a fixed vocabulary in [here](https://github.com/IBM/graph2tac/blob/bbe3a76dd0023e74c3bf076ca6f87f915d4c5a7f/graph2tac/tfgnn/dataset.py#L59).  The vocabulary for Coq names is `a-zA-Z0-9_'.` (plus unicode which we can skip for now).  Don't forget the period (`.`).
* Remove `adapt` [here](https://github.com/IBM/graph2tac/blob/bbe3a76dd0023e74c3bf076ca6f87f915d4c5a7f/graph2tac/tfgnn/dataset.py#L65) since we are no longer training the adapter if using a fixed vocab.
* Add a call to `tokenize_definition_graph` [here](https://github.com/IBM/graph2tac/blob/1cc028c46449c53fa724d90b74e85e465291a8c6/graph2tac/tfgnn/predict.py#L308) and anywhere else we are recomputing definitions.

I would be supportive of this.

jasonrute:
Where does this stand?  Is it ready to be merged?

jasonrute:
Next I'm going to benchmark this to make sure it works there.  Also, I want to wait until the names are in before merging.

PR #61, opened 2022-08-25
Add lists of important theorems

jasonrute:
Add list of theorems that we would like to use for various purposes.

Signed-off-by: Jason Rute <Jason.Rute@ibm.com>


notebooks/theorem_lists/extlib_highschoolgeom_1000.txt
```diff
@@ -0,0 +1,1000 @@
+Coq.Arith.Compare_dec.dec_ge
```
LasseBlaauwbroek, line 1: Thanks Jason! One comment: My guess is that this file has the wrong contents.

PR #77, opened 2022-09-19
Improve logging in server

jasonrute:
This PR adds a number of fixes and improvements to the G2t-server logging.
* The logs were duplicated, and this fixes that.
* TF warnings were not being caught and were being shown even if the `--tf_log_level` was set to critical.  This fixes that.
* Info-level logs no longer happen during predict steps so the logs only happen at most once per theorem.
* There is an informative summary log at the INFO level.  Here is an example
```
INFO:graph2tac:(final) Session 0, Theorem 0 summary:
  Initialization:
    Initial message size: 38400624 B
    Build network: 0.040764 s
    Update defs: 6.788260 s to update 1 def clusters
  Predict:
    Messages: 81 calls
    Ave message size: 474081.8 B/call
    Avg predict step: 0.928193 s/call
    Longest predict step: 7.700841 s on message 0
    Waiting for Coq: 0.151720 s/call
  Data and Memory:
    Total message sizes: 39.513 MB
    Current Total Memory: 17021.874 MB
    Current Physical Memory: 4322.734 MB
    Diff Total Memory: 3542.450 MB
    Diff Physical Memory: 3625.906 MB
``` 
Before I merge it:
- [ ] Test with benchmarking.  (I'm worried I turned off the ability to log messages during benchmarking.)

I left all the other info level logs a the `initialize` step.  If this is too verbose, I could make a logging level between INFO and WARN and call it maybe "summary" and put the summary message there.


LasseBlaauwbroek:
Thanks Jason. I do not understand how to use this though (wanted to suggest some simple improvements). I tried to invoke it using:
```g2t-server --arch hmodel --model . --tcp --tactic_expand_bound=8 --log_level=info --total_expand_bound=10000 --search_expand_bound=100 --update_new_definitions```
But I'm not getting any output. I'm only getting output on `verbose` and `debug` levels. But not the ouput that you show above.

**Comment by jasonrute** [comment]
@LasseBlaauwbroek No INFO messages at all?  But you see verbose messages?  That is really weird.  Can you paste your output?  If you see some INFO messages, but not my message, I forgot to mention that this message only shows when you initialize or restart Coq, so you won't see it until you go to the next theorem or restart Coq.  If it is still not working, can you paste your output and I'll debug.

**Comment by jasonrute** [comment]
Also, I didn't test the hmodel.  I'll do that later to make sure it doesn't take a different logical path through the code.

**Comment by LasseBlaauwbroek** [comment]
I'm never seeing any INFO level output. Not your summary or anything else.

**Comment by LasseBlaauwbroek** [comment]
And I cannot test the other models, because I don't have any model available to test...

**Comment by LasseBlaauwbroek** [comment]
On `main` I do see info level messages...

File: graph2tac/common.py [hunk 1]
```diff
@@ -51,6 +51,3 @@ def uuid(data_dir: Path):
 logger = logging.getLogger('graph2tac')
 logger.verbose = lambda message: logger.log(level=15, msg=message)
 logger.setLevel(logging.INFO)
-handler = logging.StreamHandler(sys.stdout)
-handler.setFormatter(logging.Formatter('%(name)s:%(levelname)s - %(message)s'))
-logger.addHandler(handler)
```
jasonrute, line 56: @LasseBlaauwbroek If it isn't working, you could try adding in back this code.  This caused duplicate messages for me, but maybe not for you.

LasseBlaauwbroek: Yes, that works.

LasseBlaauwbroek:
Some recommendations:
- I would add `msg.initialize.logAnnotation` to `Session 0, Theorem 1 summary`. That will allow us to easily crossreference a summary message to a Coq source location.
- I would add some unique prefix to every line of the summary so that we can easily grep for it.
- I would remove some of the more useless INFO messages, and try to get the other ones as part of the summary. For example, `Building network model completed in 0.000019 seconds` and `Updating definition clusters...` (with timing info) could be part of the summary.
- Consider making the messages more machine-parsable so that we can easily get summaries-of-summaries for a benchmark

LasseBlaauwbroek:
```
graph2tac:INFO - closing the server on port 33333
Traceback (most recent call last):
  File "/home/lasse/Documents/Projects/Tactician/develop/graph2tac/venv/bin/g2t-server", line 33, in <module>
    sys.exit(load_entry_point('graph2tac', 'console_scripts', 'g2t-server')())
  File "/home/lasse/Documents/Projects/Tactician/develop/graph2tac/graph2tac/loader/predict_server.py", line 870, in main
    main_loop(reader, sock, predict,
  File "/home/lasse/Documents/Projects/Tactician/develop/graph2tac/graph2tac/loader/predict_server.py", line 609, in main_loop
    logger.info(f"(final) " + log_cnts.summary_log_message())
  File "/home/lasse/Documents/Projects/Tactician/develop/graph2tac/graph2tac/loader/predict_server.py", line 300, in summary_log_message
    f"    Avg predict step: {self.t_predict/self.n_predict:.6f} s/call\n"
ZeroDivisionError: float division by zero
```

jasonrute:
Thanks for the suggestions.  Yes, I at first didn't want to mess too much with Fidel's previous code, but I agree, there isn't a need for most of those messages now.  I would still keep some of the information however (at least at some logging level), since if you are using the server manually it is nice to know what phase it is at running.  (Again, I could also move the summary up to its own logging level just a bit higher than INFO.)

Having both *short* human readable and *precise* machine readable output is definitely a challenge that I've encountered many times in this project.  I would prefer in the long run not to mix them so much, but I admit it is easier in the short run to do so.

LasseBlaauwbroek:
One simple way of making everything much more machine readable is to
1. Move the trailing SI-unit to the description
2. Change the separation between description and number to a [tab]
For example, `Avg predict step: 0.000149 s/call` becomes `Avg predict step (s/call):\t0.000149`

LasseBlaauwbroek:
Such changes will allow me to write a quick bash oneliner to get some easy summaries from a log.

jasonrute:
Ok.  I'll do that:  Next steps:
- [ ] Fix divide by 0
- [ ] Use tab format (one number per line)
- [ ] Remove superfluous logging or move to another level
- [ ] Debug why it doesn't show up on some systems and shows up duplicated on others.

jasonrute:
But also note you can only do some stats easily that way.  We could just as easily write a jsonline file or CSV and then you could do it is a shorter one-liner. :smile:

LasseBlaauwbroek:
Sure. But not much shorter, and then you potentially lose the human readable aspect...

LasseBlaauwbroek:
You don't have to keep all numbers on the same line btw. As long as you separate the columns by tabs it's fine. So you can probably make your messages more compact that way.

LasseBlaauwbroek:
LGTM!

jasonrute:
There are a lot of extra logs still, but I don't know what they mean, so I'm just going to leave them for now.

jasonrute:
Fixed some rough edges, added CPU % and a UUID, and logging the summary messages at the new "summary" level which is good for benchmarking.

PR #78, opened 2022-09-23
fix predict crash -- local & global context received together

mirefek:
Very simple bug in the end. I just forgot the 'git commit -s' again, grr... Deal with it somehow, Jason, I am finishing for today.


LasseBlaauwbroek:
This is why I love static typing :-)

jasonrute:
To be fair, while Python doesn't have static typing (and our project would benefit a lot from it), it does have a both a typing module which has good support in PyCharm and VSCode (and likely Emac with the right plugins).   We should use it.  I also has datatypes which we should be using over tuples.  This particular "tuple bug" has happened many times and would have been less likely if we have used a datatype here instead of a tuple.  I've complained about this particular piece of code for months, but let me just fix it next week so I can stop complaining and it can stop breaking our code.

PR #79, opened 2022-09-28
Use datatypes for loader outputs

jasonrute:
Now the dataloader returns proper dataclasses including `LoaderProofstate`, `LoaderAction` and `LoaderDefinition`.  This provides a uniform and easy to use interface which will hopefully reduce bugs.  I refactored all the code to use these.  Here is how they are used in practice:
* They are converted into TF (and other as follows)
  * TF2: Converts them into FlatbatchNP and FlatbatchDefNP for loading into tensorflow for training and prediction
  * TFGNN: Converts them into tuples for loading into a tensorflow dataset for training and prediction.  (The converting them back to a tuple isn't ideal, but I didn't want to bother with TF dataclasses which can require a lot of boilerplate code.  Also, they only stay a tuple for one step, before getting converted by `_make_definition_graph_tensor` to tf-gnn objects.)  Separately, the evaluation code also passes around the state and action objects seperately)
  * hmodel: Gets the information it needs from these loader classes to "train" a model or compute a prediction
* The loader objects are created in two settings
  * In the loader where they are sent by iterators for training in the 3 models
  * In the prediction server where they are created from the information being sent from Coq.

This change adds little code and in many places removes code.  Most of the added code is in doc strings (which I of course didn't need to add but thought this would be helpful since this is some of the most confusing datatypes we have in the code.)

I still need to do the following:
- [x] Test training a tf2 model 
- [x] Test training a tf-gnn model 
- [x] Test training a hmodel model 
- [x] Test server with tf2 model making sure it still can prove trivial theorems on first prediction using existing model
- [x] Test server with tf-gnn model making sure it still can prove trivial theorems on first prediction using existing model
- [x] Test server with hmodel making sure it still can prove trivial theorems on first prediction using existing model
- [ ] Modify the `tfgnn.ipynb` notebook accordingly.

### DIFF
--- a/graph2tac/loader/data_server.py
+++ b/graph2tac/loader/data_server.py
@@ -3,7 +3,8 @@ data server provides the class to serve data to training
  - from a collection of bin files in a filesystem
  - from messages in a stream in interactive session (to be implemented)
 """
-from typing import List, Tuple, Callable, NewType
+from typing import Iterable, List, Tuple, Callable
+from numpy.typing import NDArray
 
 import time
 import os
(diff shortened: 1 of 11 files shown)


jasonrute:
I'm going to handle the notebook in a separate PR.  Everything else seems to work fine.

PR #82, opened 2022-10-13
Speed up prediction

jasonrute:
Speeds up prediction including the definition task, by removing the need for `tf.Dataset` and by compiling most of the code.


graph2tac/tfgnn/predict.py
```diff
@@ -17,6 +17,59 @@
 from graph2tac.predict import Predict, predict_api_debugging, cartesian_product, NUMPY_NDIM_LIMIT
 
 
+def stack_dicts_with(f, ds):
```
jasonrute, line 20: Is this a bad place for this code?

jasonrute:
I changed my mind.  Next week I'll fix up all the duplicated code in this PR, but we need to merge it since it is important for benchmarking.

PR #85, opened 2022-11-10
Move v13 into main

jasonrute:
This makes v13 the main data format.  It also brings in a preliminary version of the python loader.  It is not the default, but it is available to the hmodel training process.  (I tested it and it works on the hmodel training.)

Also the dependencies are now fixed in the setup.py so one only needs to do `pip install .` (or `pip install -e .` for development).


mirefek (approved):

pestun (commented):
looks good

pestun (approved):

PR #86, opened 2022-11-13
Vasily/v13+python loader

pestun:


tests/mini_stdlib/test0/fingerprint.out
```diff
@@ -1 +1 @@
-9.4876E-02 0.0000E+00 6.9364E-02 3.9104E+00 4.3959E+00 3.9078E+00 3.0695E-01 0.0000E+00 7.5145E-02 3.9042E+00 4.3903E+00 3.9026E+00
```
jasonrute, line 1: Why did the test results changed?  Is it because the data order changed with loader changes?

pestun: Yes we changed fingerpints in much earlier commit: https://github.com/IBM/graph2tac/pull/86/commits/e3116eee0c5049d1d5a69e4dffc42fdf793ca9a4

The purpose was to align tactic hash indexing between the two loaders. 

The original loader was indexing tactic hashes by the lexicographical order (by value of the tactic hash).

Mirek's loader was using the chronological order of walking through the dataset, and I switched to the chronological order as well so that it would be easier for Mirek to compare the output of the two loaders. 

With that change the two loaders seem to use the same ordering of tactic hashes.

jasonrute:
Also, it would be good to add an explicit reference to the version of `coq-tactician-reinforce` you are using in the `setup.py` file (like it is in `main`).  You don't need to do that for this specific PR.  When it is merged, the current version in `main` will stay the same since you didn't touch the setup.py file.   However, it would be good in general to make our dependencies explicit in the branches to avoid future issues.  So I'd recommend after this is merged, to merge/rebase `main` into the current outstanding feature branches and change the `coq-tactician-reinforce` version accordingly.

PR #87, opened 2022-11-17
Mirek/cython speedup

mirefek:
Cython loader set as default for tf2 / tfgnn, setup.py updated


jasonrute:
Let me try to run it (setting up my conda environment as needed) and then if I can do that, I'll merge it.

graph2tac/tf2/train.py
```diff
@@ -20,13 +20,15 @@
 from dataclasses_json import dataclass_json
 import yaml
 
+import graph2tac.loader.py_data_server   # import capnp before tensorflow
```
jasonrute, line 23: Just so I know, why is it important to import capnp before tensorflow?  (Or is it just that tensorflow should be delayed since it takes so long to import?)

setup.py
```diff
@@ -115,6 +115,7 @@ def run(self):
         'dataclasses-json',
         'pyyaml',
         'graphviz',
-        'pytact @ git+https://git@github.com/coq-tactician/coq-tactician-reinforce#f6992c129ad99aa47fb80a6322f6ead5e86bda92'
+        #'pytact @ git+https://git@github.com/coq-tactician/coq-tactician-reinforce#f6992c129ad99aa47fb80a6322f6ead5e86bda92'
```
jasonrute, line 118: Can you delete this line?  I think it is a bad habit to comment out code like this.  We can always use git blame and the file history to figure out what this used to be.

graph2tac/tf2/train.py
mirefek: On Cholesky cluster, it crashes otherwise (for not being able to find the linked libraries). I don't understand it properly, just know that it is not the first time when something had to be loaded before tensorflow.

setup.py
mirefek: done

graph2tac/tf2/train.py
jasonrute: Have you tested the predict server?  Do we have to do the same there?

jasonrute: Oh, that probably already loads capnp before TF actually.

jasonrute:
`g2t-train-hmodel --help` gives the error:
```
import pytact.graph_api_capnp as graph_api_capnp
ModuleNotFoundError: No module named 'pytact.graph_api_capnp'
```
But I don't get this error in `main`.  Are we sure we are using the right branch of coq-tactician-reinforce?

jasonrute:
It looks like that commit of coq-tactician-reinforce doesn't have `pytact` anymore in setup.py.  (https://github.com/coq-tactician/coq-tactician-reinforce/compare/f6992c129ad99aa47fb80a6322f6ead5e86bda92...5d3f779a76ff64023816ef0fb908afc71dd46cc8#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7) @LasseBlaauwbroek?

LasseBlaauwbroek:
I switched to a `pyproject.toml` file in the cython branch. Apparently this is the future of Python packaging (and I didn't know how to do something in the old way so I just switched).

jasonrute:
@pestun Ok, makes sense.  How do we specify the dependency in our `setup.py`, and what should the import be in the python code?

LasseBlaauwbroek:
As far as I know it is the same as it has always been. The error you're getting suggests that something may not have been installed correctly.

LasseBlaauwbroek:
Pip is famous for not doing dependency solving, so it might have screwed up the dependencies. Did you try installing in a clean virtualenv?

graph2tac/loader/py_data_server.py
```diff
@@ -1,5 +1,4 @@
 from pathlib import Path
-import capnp
```
LasseBlaauwbroek, line 2: I can confirm @jasonrute's error. It seems to be caused by the removal of this line.

mirefek: ok, if it helps to put the line back, I can

jasonrute: @mirefek Can you put it right above `import pytact.graph_api_capnp as graph_api_capnp` and make a comment that this line is needed for the next import to work?

jasonrute: Linting tools will show it as a dead import, so it is good to have a comment to know why it is there.

LasseBlaauwbroek: Yeah it looks like I was mistaken about the necessity. That being said, if you omit it you are implictly relying on other libraries to import it... That's somewhat fragile.

jasonrute:
@mirefek I can close this PR right?  It is part of the other py server PR?

mirefek:
Yes, I don't remember this branch too well but git says it is contained in py_predict_server, and it also looks so by looking at the code.

PR #88, opened 2022-11-22
Load datasets faster and without bugs

jasonrute:
This PR fixes the issues with loading the datasets where we were caching at the wrong places.  Now the datasets should be shuffled between epochs and loading validation and training together will only require one pass through the dataset.  I've also turned off computing statistics since it was so time consuming and had its own inefficiencies.

### DIFF
--- a/graph2tac/tfgnn/dataset.py
+++ b/graph2tac/tfgnn/dataset.py
@@ -106,6 +106,7 @@ class Dataset:
 
         # apply the symmetrization and self-edge transformations
         proofstate_dataset = proofstate_dataset.apply(self._preprocess)
+        proofstate_dataset = proofstate_dataset.cache()
 
         # create dataset of split labels
         split_logits = tf.math.log(tf.constant(split, dtype=tf.float32) / sum(split))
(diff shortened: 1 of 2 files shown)


jasonrute:
I still want to double check that this is working correct under the hood before merging.

jasonrute:
I added a fix to the datasets.  I also double checked with some print statements.  (1) The shuffling is happening correctly.  (2) It is processing each proof state and each definition only once.

jasonrute:
Some statistics for running 2 epochs on 4 cores and 2 GPUs with a 90% train, 10% validation split.
* before
  * time: 91m47s (a decent amount of this time is the statistics that I turned off)
  * First epoch (train only): 1457s
  * First epoch (train+validate): 2555s (this is extra long because of all the duplicate dataset loading in the validation phase)
  * Second epoch (train+validate): 1309s (this should be typical of later epochs, but it does seem longer than I expected)
  * Max memory (virtual): 166G 
  * Max memory (physical): 74.3G
* after
  * time: 37m14s
  * First epoch (train only): 1396s
  * First epoch (train+validate): 1489s
  * Second epoch (train+validate): 664s
  * Max memory (virtual): 140G
  * Max memory (physical): 47.8G
  
Overall, this looks much better.  I still don't understand why the second epoch is longer in the "before" than the "after" since my changes shouldn't effect that.  It isn't exactly in the logs, but I think the "before" spent a lot longer on the validation part of the second epoch than the "after".  I'm doubtful this will end up say doubling the speed of training.  Anyway, whatever the reason, this looks to be a good fix overall!

jasonrute:
Mirek said he is happy with these changes, so I'm going to go ahead an merge them now.

PR #89, opened 2022-11-22
Improve tfgnn memory

jasonrute:
This reducing the memory needs of this one part of code and fixes a ~memory leak~ a place in the code which can cause OOM errors on the GPU.

Initial code from Mirek.  Cleaned up by me.

Signed-off-by: Jason Rute <Jason.Rute@ibm.com>


jasonrute:
Fidel also suggested another way to do this that is essentially the same with different commands.  I'm fine with this way since I had already written the code.

jasonrute:
I also checked that the results are an exact match to the previous code.

mirefek (approved):
lgtm

PR #90, opened 2023-01-18
Jrute/model improvements

jasonrute:
This PR updates the model with a few basic fixes including the following:
* Dropout is now before the residual connection
* There is an optional FFN block after the convolution and before the dropout/residual.  This can be specified in the parameter file.  Like elsewhere in Fidel's code, the final layer of the FFN which must project back down to the hidden size is hard coded into the code, so e.g. a two layer FFN (like in a transformer) is specified with just the first layer.  (Also note that this last layer was there already in the code because of the way residual connections work in TFGNN, we need to project down to the hidden dimension.) 
* There is an option to "sum_loss_over_tactic" which works as follows:
  * The loss of the tactic "<base_tactic> <local_arg> <global_arg>" is now the loss of the three parts added together.  The loss of the batch is aggregated automatically by Keras. This is technically implemented by returning a loss array of shape `[batch]` for each of tactic, local_args, global_args.  Since they all have the same shape, it is equivalent to adding the losses element wise before reducing the losses to a single number.
  * If that option is off, then it will behave like before. The code is different since it was better to implement both options with the same code and only change it at the end, but it should behave the same.  With that option set to false, it will return an array of losses for each local argument in the batch and for each global in the batch (and keras will take the mean of each array before adding them together).  Since the number of locals and globals is different for each batch, the relative effect of a argument will change from batch to batch.  (Also, since there are usually more tactics than arguments of a given type, the tactics will be weighted less than the args, which is why previously we needed to lower the loss coeff of the arguments.)
* I've updated the default parameters of the tfgnn.  Here are some highlights:
  * Turned on "sum_loss_over_tactics" and changed argument loss coefficient to 1.0.  This works very well in my experiments.
  * Added a FFN layer.
  * Other changes to the network hyperparameters that work well after extensive testing.

I've tested this (without the definition loss) on ministdlib and it does very well.  I've also tested a similar set of settings and architectural fixes on all of stdlib (in a different branch) and it also does well.

### DIFF
--- a/graph2tac/tfgnn/default_global_argument_prediction.yml
+++ b/graph2tac/tfgnn/default_global_argument_prediction.yml
@@ -5,8 +5,11 @@ gnn_type: simple_convolution_gnn
 gnn_config:
   hops: 8
   dense_activation: relu
-  residual_activation: relu
+  ffn_layers:
+  - activation: relu
+    units: 256
   dropout_rate: 0.1
+  residual_activation: null
   layer_norm: true
   reduce_type: max
   final_reduce_type: max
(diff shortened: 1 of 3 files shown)


mirefek (approved):
Sounds reasonable

PR #91, opened 2023-01-26
Improve definition task and global argument prediction

jasonrute:
These are improvements to the model definitions and global argument prediction:
* (Optional) unit normalize embeddings.  I do this correctly by adding it as a constraint to the Embedding.  If I just normalize the embeddings when I use them, then the layer normalization brings the embeddings to 0.0 and eventually I start dividing by zero.  This keeps them in place.
* (Optional) Add cosine similarity as source of logits for global argument prediction.  This includes normalizing the key and query and also dividing the cosine similarity by a learned temperature parameter which is common.  Note, I chose to normalize the embeddings since we may want to try cosine similarity with embeddings which are not normalized.  I guess a not too difficult future optimization would be when creating the LogitsFromEmbedding object to add a parameter stating if the embeddings are already normalized.
* Add local and global arg heads.  This way the local and globals get there own keys.  They are just dense layers that branch off a common argument head.
* Change definition loss to use norm squared instead of mean squared error.  It is equivalent, but isn't effected so much by hidden layer size.
* Change default final reductions to mean instead of max.  Changing the final reduction from max to mean makes a huge improvement.  Max seems to make the definition task and tactic task compete against each other.  As one gets better, the other gets worse.  It eventually overcomes this competition, but it takes a long time.  It seems better with mean.
* I also changed the reduction after each convolution from max to mean.  I don't know if this is better, and I'm still experimenting a bit, but in principle we have some graphs with multi-edges, and max can't distinguish a multi-edge from a single edge.
* Increase definition task to 500.  It previously was 100.0 using mean squared error, which would be 100/128 = 0.78 using the norm squared loss.  So we are increasing it by over 500x.  This is made possible with the above improvements.

There also are a few internal changes to the predict code:
* I moved a lot more of the functions related to embeddings (like enlarging the embedding table) into the embedding class itself. The reason was that as I made changes to the embedding object it was hard to keep track of where else it was used in the code and if I needed to make changes there.  It will also help facilitate future experiments where I replace the embeddings in the code between epochs.
* Similarly, I moved some of the functionality of the LogitsFromEmbedding class back into the class.

More notes:
* I haven't merged the previous PR this is built on top of so this PR is a mix of both.  It would be easiest to just look at the most recent commit.
* I still need to test that the predict server works correctly.
* I also want to do a full scale run of this model.

### DIFF
--- a/graph2tac/tfgnn/default_global_argument_prediction.yml
+++ b/graph2tac/tfgnn/default_global_argument_prediction.yml
@@ -24,7 +24,7 @@ arguments_head_config:
   activation: linear
   recurrent_activation: relu
   recurrent_depth: 2
-global_cosign_similarity: true
+global_cosine_similarity: true
 arguments_loss_coefficient: 1.0
 sum_loss_over_tactic: true
 dynamic_global_context: true
--- a/graph2tac/tfgnn/default_global_argument_prediction.yml
+++ b/graph2tac/tfgnn/default_global_argument_prediction.yml
@@ -1,5 +1,6 @@
 prediction_task_type: global_argument_prediction
 hidden_size: 128
+unit_norm_embs: true
 tactic_embedding_size: 128
 gnn_type: simple_convolution_gnn
 gnn_config:
@@ -11,8 +12,8 @@ gnn_config:
   dropout_rate: 0.1
   residual_activation: null
   layer_norm: true
-  reduce_type: max
-  final_reduce_type: max
+  reduce_type: mean
+  final_reduce_type: mean
 tactic_head_type: dense_tactic
 tactic_head_config:
   hidden_layers:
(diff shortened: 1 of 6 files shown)


graph2tac/tfgnn/default_global_argument_prediction.yml
```diff
@@ -1,24 +1,30 @@
...
     units: 256
 arguments_head_type: simple_rnn
 arguments_head_config:
-  activation: relu
-  recurrent_activation: tanh
+  activation: linear
+  recurrent_activation: relu
   recurrent_depth: 2
-arguments_loss_coefficient: 0.1
+global_cosign_similarity: true
```
jasonrute, line 27: Mirek noticed spelling error.

jasonrute: Fixed.

jasonrute:
I walked Mirek through the code and he gave his approval as long as I checked that it still works in the server.  I fixed a bug which prevented the saving all the weights to checkpoints.  I also used prints to ensure that the embedding table was still being updated correctly in the server.  I also added an initializer to the embeddings so even without training they are on the unit sphere.  It's ready to merge.

PR #93, opened 2023-01-30
Fix bug in def name embedding

jasonrute:
Each definition name is a different length.  There was a line in the code which I'm deleting in this PR, which padded the name to be the same size as the longest name in the batch.  There was no masking on the padding (and it happened after the token embeddings, so it padded the input to the RNN model with zero vectors).

Every batch computed the name slightly differently based on the padding, and the embeddings at inference time (which were not padded since the batches have size one) came out differently than at training time (where for most names there was at least some padding).  This PR removes any such padding since RNN models in TF don't need to be padded like this.


jasonrute:
The model runs.  The definition names now come out the same at inference and training time, and I've already discussed this with Mirek, so I'm going to go ahead and merge it.

PR #95, opened 2023-02-01
Expand shuffle buffer to shuffle all the data

jasonrute:
The TFGNN network uses the TF Dataset pipeline to process all the data.  One part of this processing is shuffling the data for training.  Since the dataset (at least how we are using it) is built around a python iterator, it doesn't know how many datapoints are in an epoch and doesn't generate a whole epoch's worth of data at once.  To shuffle, one gives the shuffle function a buffer size.  Our buffer size was 10e4, which was much smaller than the dataset size of 5e5 (half a million datapoints).  This actually was leading to big problems.  The data was only locally shuffled and by the end of an epoch it seems the network forgot about the data at the start of the epoch.  For example, near the end of training the model had a training accuracy of about 50% at the start of the epoch, but 70% across the whole epoch.

This PR increases the buffer size to the point where we are shuffling all the data, and makes sure any shuffles are before repeat so that we don't buffer any data more than once.  Since we are already caching all the data, this doesn't effect the amount of memory used substantially.

**I've just tested it on 10 epochs and the network learns substantially faster with this change. It reaches a validation accuracy of 50% in 10 epochs whereas it took four times longer previously to reach that accuracy.**

As we expand our dataset we are going to have to figure out what to do as the dataset grows in size.  But I'd like to merge this change for now as we figure that out.


LasseBlaauwbroek:
Ha, that is amazing, of all the things that could be going wrong...

> As we expand our dataset we are going to have to figure out what to do as the dataset grows in size.

I believe that Mirek is reworking the dataset splitting infrastructure. I'd imagine this would include modifying this code? So this might not be such a big problem.

mirefek:
Hm, these Datasets seem to be done in somewhat impractical way... I have two questions:

1. Why exactly do we need this SHUFFLE_BUFFER_SIZE number? What happens if we don't provide it? Guessing a random number bigger than the dataset doesn't seem as a very conceptual solution.
2. As a preliminary solution, you could also shuffle the data at the beginning by the loader by setting `shuffled = true` here: https://github.com/IBM/graph2tac/blob/247aa3ceb15c96eecf19cf1b5051c7a71a11edc3/graph2tac/tfgnn/dataset.py#L800 Then the order is the same every epoch but at least random.

jasonrute:
> I believe that Mirek is reworking the dataset splitting infrastructure. I'd imagine this would include modifying this code? So this might not be such a big problem.

Yes.  That is how I see it.

jasonrute:
> 1. Why exactly do we need this SHUFFLE_BUFFER_SIZE number? What happens if we don't provide it? Guessing a random number bigger than the dataset doesn't seem as a very conceptual solution.

It is a required parameter (you can't pass in `None` either).  From the docs:
> For perfect shuffling, a buffer size greater than or equal to the full size of the dataset is required.

> 2. As a preliminary solution, you could also shuffle the data at the beginning by the loader by setting `shuffled = true` here: https://github.com/IBM/graph2tac/blob/247aa3ceb15c96eecf19cf1b5051c7a71a11edc3/graph2tac/tfgnn/dataset.py#L800
>     Then the order is the same every epoch but at least random.
We could try this, but I'd want at least to have some shuffling locally with a buffer of say 10e4.  I think having constant batches would be detrimental.  However, even with that, I'm worried it won't be as random as it needs to be.  If you find this temporary solution unpalatable, I can try the other approach.

PR #97, opened 2023-02-01
Fix bug in predict embeddings

jasonrute:
I made a mistake when refactoring the embeddings.  I tested it but not well enough.  (I tested that the embedding table was being extended properly, but apparently it wasn't being rebuilt since I extended it in-place instead of making a new layer.)  This fixes it.  I've now tested that an entire benchmark works.


jasonrute:
This is a really small bugfix to some previous recent changes, so I'm just going to merge this.

PR #98, opened 2023-02-02
remove C loader and TFRecord

mirefek:
I also removed the TFRecord but am not focusing on it yet. Rather doing a simple step so that Jason can continue with removing TF2 (since I modified also other files such as setup.py).


setup.py
```diff
@@ -91,13 +22,8 @@ def run(self):
     package_data={'graph2tac.loader': [f'clib/graph_api_v{CAPNP_V}.capnp']},
```
LasseBlaauwbroek, line 22:
```suggestion
```
Can also be removed I think.

setup.py
```diff
@@ -1,88 +1,19 @@
-from pathlib import Path
 import sys
 import setuptools
 from setuptools import find_packages, setup
-from setuptools.command.build_ext import build_ext
-import subprocess
 
 # These next three lines fetch and import numpy, which is needed for installation
 import setuptools.dist
 setuptools.dist.Distribution().fetch_build_eggs(['Cython>=0.15.1', 'numpy>=1.10'])
```
LasseBlaauwbroek, line 7:
```suggestion
```
Can be deleted.

setup.py
```diff
@@ -1,88 +1,19 @@
...
-import numpy
-
 
 # please indent the lists with one element per line
 # for convenience of version control tools
 
 CAPNP_V=14
```
LasseBlaauwbroek, line 12:
```suggestion
```
Can be deleted.

LasseBlaauwbroek (approved):
Some minor additional suggestions for deleting things. Otherwise things look good (assuming that it still runs of course, I didn't test that).

LasseBlaauwbroek:
Removing 9000 lines of code feels pretty good :-)

setup.py
mirefek: resolved with https://github.com/IBM/graph2tac/pull/98/commits/34c9863b7f235a1d59a219edd880fe7a74e15295

graph2tac/loader/py_data_server.py
```diff
@@ -199,14 +198,16 @@ def __init__(self,
...
-        if fname_order_key is not None:
-            fnames.sort(key = fname_order_key)
+        # fnames = [
+        #     fname.relative_to(data_dir)
+        #     for fname in top_sort_dataset_modules(data_dir)
+        # ]
+        # assert set(fnames) == set(self._data.keys())
+        # if fname_order_key is not None:
+        #     fnames.sort(key = fname_order_key)
+
```
jasonrute, line 210: Let's delete this all this commented out code.

setup.py
```diff
@@ -1,103 +1,21 @@
...
 import sys
-import setuptools
 from setuptools import find_packages, setup
-from setuptools.command.build_ext import build_ext
-import subprocess
-
-# These next three lines fetch and import numpy, which is needed for installation
-import setuptools.dist
-setuptools.dist.Distribution().fetch_build_eggs(['Cython>=0.15.1', 'numpy>=1.10'])
-import numpy
```
jasonrute, line 11: I'm so happy we are removing this.  It has given me so many issues!

setup.py
```diff
@@ -1,103 +1,21 @@
...
     entry_points={'console_scripts':
                   [
-                      'g2t-preprocess=graph2tac.loader.preprocess:main',
                       'g2t-train=graph2tac.tf2.train:main',
                       'g2t-train-tfgnn=graph2tac.tfgnn.train:main',
-                      'g2t-stat=graph2tac.loader.graph_stat:main',
-                      'g2t-printlog=graph2tac.printlog:main',
-                      'g2t-unpack=graph2tac.loader.unpack:main',
-                      'g2t-server=graph2tac.loader.predict_server:main',
                       'g2t-py-server=graph2tac.loader.py_predict_server:main',
```
jasonrute, line 19: Should we just go back to calling this `g2t-server`?

jasonrute (approved):
Looks good!  I made a few minor suggestions.

setup.py
mirefek: fine, then we could also remove the prefix from py_..._server

LasseBlaauwbroek: I'm not against this, but as a warning: Renaming the file might get you some nasty conflicts in other branches. Better to wait with renaming the file until everything is merged.

mirefek: I see because git will think it is the original file... I think my `split_options2` will be fine, they have already branched from `simplify`. So the question is about @jasonrute 's WIP (about definition recalculation) which we are not sure if / when we are actually going to merge. Perhaps, Jason, you could first merge `simplify` into your branch, and then I could do the renamings... would it be fine like that? Or is it another issue? I am not sure about how git does merging.

LasseBlaauwbroek: It's not necessarily about the original file. Even without that, you might run into issues because git's merge heuristics are just not very good across different file names. I'd just not take that risk just to remove `py` from a filename...

mirefek: Fine, I will create another branch, and try to merge `simplify` to that, what happens

LasseBlaauwbroek:
Great! Let's merge (I cannot press the button because I'm too lazy to set up 2FA).

mirefek:
I pressed the button :-)

setup.py
mirefek: https://github.com/IBM/graph2tac/pull/101

PR #99, opened 2023-02-03
Mirek/split options2

mirefek:


LasseBlaauwbroek:
Can you re-target this PR to the `main` branch? It's still showing all of the changes from the previous PR. Github does not automatically update that (very annoying). There should be some option near the top of the PR to retarget.

mirefek:
I will try, what is the command for retargetting?

LasseBlaauwbroek:
> There should be some option near the top of the PR to retarget.

Or you can just close and re-open.

PR #100, opened 2023-02-03
Mirek/split options2

mirefek:


graph2tac/loader/py_data_server.py
```diff
@@ -171,23 +171,67 @@ def cluster_to_graph(self, cluster):
...
+        return 0
+    def cluster(self, d : list[Definition]):
+        return self.datapoint(d[0])
+
 class DataServer(AbstractDataServer):
     def __init__(self,
                  data_dir: Path,
                  max_subgraph_size,
-                 split: tuple[int, int, int] = (8, 1, 1),
+                 split = (8, 1, 1),
```
LasseBlaauwbroek, line 215: Isn't it time for us to get rid of the concept of holdout-set? I think we have all agreed in the past that we will almost certainly not use this.

mirefek:
https://github.com/IBM/graph2tac/pull/100 Merged without issues

graph2tac/loader/py_data_server.py
```diff
@@ -171,23 +171,67 @@ def cluster_to_graph(self, cluster):
...
+        return get_split_label(ident_64, self.split, self.random_seed)
+    def cluster(self, d : list[Definition]):
+        return 0
+
+class SplitByFilePrefix(Splitter):
+    def __init__(self, *prefixes_per_label):
+        self.data_server = None # to be set by data_server
+        self.prefixes_per_label = []
+        for label, prefixes in enumerate(prefixes_per_label):
+            if isinstance(prefixes, str):
```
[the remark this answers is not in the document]
mirefek: I just want the config being flexible, where we parse the config is up to discussion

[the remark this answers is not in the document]
LasseBlaauwbroek: In my opinion, you should always determine the shape or your data as early as possible. So I'd do it when parsing the yaml file.

graph2tac/loader/py_data_server.py
mirefek: Up to discussion. I don't mind too much allowing more than two split labels.

graph2tac/loader/py_data_server.py
```diff
@@ -171,23 +171,67 @@ def cluster_to_graph(self, cluster):
...
+                prefixes = [prefixes]
+            else:
+                prefixes = list(prefixes)
+            self.prefixes_per_label.append((label+1, prefixes))
+    def datapoint(self, d : Definition):
+        fname = self.data_server.graphid_to_fname[d.node.graph]
```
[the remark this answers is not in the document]
mirefek: In case there would be another Splitter taking something else from the data_server

graph2tac/loader/py_data_server.py
LasseBlaauwbroek: I would at least change the default to `(9, 1, 0)`.

graph2tac/loader/py_data_server.py
```diff
@@ -171,23 +171,67 @@ def cluster_to_graph(self, cluster):
...
 
+        if isinstance(split, tuple) and isinstance(split[0], int):
+            self.split = SplitByHash(split, split_random_seed)
+        elif isinstance(split, tuple) and isinstance(split[0], (str, list)) and isinstance(split[0][0], str):
+            self.split = SplitByFilePrefix(split)
+        elif isinstance(split, Splitter):
+            self.split = split # ignore split_random_seed
+        else:
+            raise Exception("Cannot interpret split:", split)
+        self.split.data_server = self
```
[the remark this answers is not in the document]
mirefek: Passing directly a Splitter is among the options. I agree I am sort of lazy / hacky here with keeping backward compatibility. At some point, we need to parse the config file, and create the appropriate Splitter. Now, I am doing it here but I agree there could be a better option.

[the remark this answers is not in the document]
mirefek: For example, TF2 is using the old split argument, and I am not sure if we are already fully getting rid of it.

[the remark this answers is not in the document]
LasseBlaauwbroek: Yes, we are getting rid of TF2 (Jason should be working on it). I think that the time for keeping backwards compatibility is over. I would just change it to what we think is good code.

LasseBlaauwbroek:
As a security measure, I think it might be a good idea to put in a check that all of the prefixes are self-consistent. That is, none of the files in the training set should reference the validation set. Otherwise we should error out.

mirefek:
Good idea, I will implement some checking

graph2tac/loader/data_server.py
```diff
@@ -171,23 +171,67 @@ def cluster_to_graph(self, cluster):
...
+    def __init__(self, split = (8,1,1), random_seed = 0):
+        self.data_server = None # to be set by data_server
+        self.split = split
+        self.random_seed = random_seed
+    def datapoint(self, d : Definition):
+        # to make it identical to vasily's loader
+        ident_64 = np.array(d.node.identity, dtype = np.uint64).item()
+        return get_split_label(ident_64, self.split, self.random_seed)
+    def cluster(self, d : list[Definition]):
+        return 0
```
LasseBlaauwbroek, line 190: Why are clusters always part of training?

mirefek: That is the old behavior

LasseBlaauwbroek: Was there an actual historical reason for this, or is this just a bug? If the latter, I'd fix it...

mirefek: to me, it is a logical behavior

mirefek: when you randomly split on lemmata

mirefek: what other split on definitions do you imagine?

LasseBlaauwbroek: Ah, okay. I misunderstood the purpose of `datapoint` and `cluster`. I get what is happening now, but I was interpreting them as doing exactly the same, except that one took a list. Perhaps rename them to `lemma` and `definition_cluster`?

LasseBlaauwbroek:
Question: Currently you are keeping a mapping from graphid to filenames. Then, for every datapoint, you lookup its graphid, translate to the filename and try to decide in which split the file is. Wouldn't it be simpler and faster, to pre-calculate for each graphid to which split it belongs? Then, deciding where a datapoint belongs is a question of one map-lookup.

mirefek:
Good point, I originally went for the easy fix, I will rewrite it this way.

graph2tac/tfgnn/dataset.py
```diff
@@ -493,16 +477,16 @@ class DataServerDataset(Dataset):
...
         @param data_dir: the directory containing the data
         @param max_subgraph_size: the maximum size of the returned sub-graphs
         @param kwargs: additional keyword arguments are passed on to the parent class
         """
         self.data_server = DataServer(data_dir=data_dir,
                                       max_subgraph_size=max_subgraph_size,
-                                      split=(1,0,0),
-                                      split_random_seed=0)
+                                      split = tuple(split),
+        )
```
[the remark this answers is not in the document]
jasonrute: But also see my proposal where we package the random seed inside the split.

graph2tac/tfgnn/dataset.py
```diff
@@ -493,16 +477,16 @@ class DataServerDataset(Dataset):
     proofstate_data_spec = (state_spec, action_spec, graph_id_spec)
     definition_data_spec = (loader_graph_spec, num_definitions_spec, definition_names_spec)
 
-    def __init__(self, data_dir: Path, max_subgraph_size: int = 1024, **kwargs):
+    def __init__(self, data_dir: Path, split = None, split_random_seed = 0, max_subgraph_size: int = 1024, **kwargs):
```
jasonrute, line 480: Again, I would removed the defaults.  In particular, `split = None` would likely crash the code, right???

graph2tac/loader/py_data_server.py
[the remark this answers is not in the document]
jasonrute: See my suggestion below.  And yes, I'm getting rid of TF2 this week hopefully.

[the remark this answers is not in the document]
jasonrute: Actually, I'll just repeat it.  We can use this pattern often used by Fidel in the yaml files:
```
split_method: hash
split:
  proportions:
  - 0
  - 1
  split_random_seed: 0
```
So then we pass that to the DataServer as `split_method: str` and `split: dict`.  This will make it easy to add more split methods as needed.

jasonrute:
Also, when you merge in main, which now has tests, please update the tests (both the parameter files and results) and add a test for splitting on prefix.  You can add the test in one of the datasets.  (I want to see if my README in `tests/README.md` is good, so first look there to see if you can figure out how to do all this with the tests.  If not, I'm happy to help and we can update the README.)

graph2tac/loader/data_server.py
mirefek: I wouldn't `split.lemma_split` feels having unnecessary repetition to me.

graph2tac/tfgnn/dataset.py
```diff
@@ -172,25 +158,23 @@ def graph_constants(self) -> GraphConstants:
             graph_constants.edge_label_num += 1
         return graph_constants
 
-    def stats(self, split: Tuple[int, int] = (9,1), split_random_seed: int = 0) -> dict[str, dict[str, int]]:
+    def stats(self) -> dict[str, dict[str, int]]:
         """
         Compute statistics for the proof-state and definition datasets.
 
         @param split: a pair of (not necessarily normalized) probabilities for the train and validation splits
         @param split_random_seed: the seed to use for the train/validation split
```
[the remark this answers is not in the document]
mirefek: What is wrong?

graph2tac/tfgnn/dataset.py
```diff
@@ -660,15 +644,14 @@ def _loader_to_proofstate_data(loader_data: tuple[LoaderProofstate, LoaderAction
 
         return (state_tuple, action_tuple, id)
 
-    def _proofstates(self) -> tf.data.Dataset:
+    def _proofstates(self, label : int = 0) -> tf.data.Dataset:
```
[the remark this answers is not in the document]
mirefek: Maybe in case you don't do any split, so you don't want to think about them?

graph2tac/loader/data_server.py
```diff
@@ -407,35 +462,34 @@ def _datapoint_text(self, i):
...
-            if get_split_label(
-                    np.array(d.node.identity, dtype = np.uint64).item(),
-                    self.split, self.split_random_seed
-            ) == label
+            if self.split.datapoint(d) == label
         ]
 
-    def data_train(self, shuffled: bool = False,  as_text: bool = False):
-        train_ids = self._select_data_points(0)
+    def get_datapoints(self, label : int, shuffled: bool = False, as_text: bool = False):
```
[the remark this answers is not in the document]
mirefek: I would keep these further renaming ideas for the upcoming "dataset pipeline simplification", first I just wanted to reasonably add the splits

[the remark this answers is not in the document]
mirefek: signature is fine

graph2tac/loader/data_server.py
jasonrute: That is a fine point.  We can leave it as `lemma` and `definition cluster`.

graph2tac/loader/data_server.py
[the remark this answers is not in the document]
jasonrute: I meant the output signature which is missing.  Although I realize it is more complicated than I thought.  I think it is something like `-> tuple[str, str] | tuple[LoaderProofstate, LoaderAction, int]`.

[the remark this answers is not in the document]
jasonrute: Yes, it is fine to delay other simplications.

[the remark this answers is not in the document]
mirefek: I added the signature (hopefully right)

graph2tac/tfgnn/dataset.py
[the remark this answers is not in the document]
jasonrute: The doc string mentions `@param split_random_seed`, `@param split_random_seed` which are now removed.

mirefek:
I believe that with the last two commits, I addressed your comments, @LasseBlaauwbroek , @jasonrute . In particular more "static" typing, and consistency checks for the prefix labels. See if you have any further comments, or there is something I missed.

graph2tac/loader/data_server.py
```diff
@@ -171,21 +172,95 @@ def cluster_to_graph(self, cluster):
...
+        # conversion from graph indices to filenames
+        graphid_to_fname = {
+            file_data.graph : str(fname)
+            for fname, file_data in data_server._data.items()
+        }
+        assert set(graphid_to_fname.keys()) == set(range(len(data_server._data)))
+        graphid_to_fname = [
+            graphid_to_fname[i]
+            for i in range(len(data_server._data))
+        ]
```
LasseBlaauwbroek, line 219: Is all of this really needed? As far as I can tell, you never actually look something up in these maps. You only use the domain of the map.

LasseBlaauwbroek (approved):
One final comment. Otherwise, assuming that this works, I propose to merge. There are still improvements to be made, but I agree with Mirek that this can be done as part of the bigger effort to simplify the data-pipeline.

graph2tac/loader/data_server.py
mirefek: It is a way to ensure that every graphid `0..len(data_server._data)-1` occurs there exactly once.

mirefek:
The tests are not giving the preset values so far. I already had a discrepancy with Jason also before, now it seems completely different -- it can be due to a different way of splitting (so possibly differently shuffled) but maybe, I should look at it more carefully.

mirefek:
And @jasonrute also wanted something more with the tests -- add there an example of the prefix split, right?

jasonrute:
Yes, if you could add one test (under one of the datasets) that would be helpful.

graph2tac/loader/data_server.py
LasseBlaauwbroek: You mean that you want to check that there is a bijection between filenames and graphid's? Okay, fine I guess, although if we want to test this, it might be better to do that centrally in pytact. (Although it seems slightly redundant, because in pytact essentially the opposite computation is happening to the one here...)

jasonrute:
As for getting different numbers in the tests, since you changed the splits, I imagine it will change the numbers.  We are now splitting by lemma instead of proof step so we will have different train-valid split.  Also, when you removed the split from the Dataset pipeline, I'm sure it changed the randomness.  (We aren't using Jax, so we don't have a lot of control over the flow of the random number generator unfortunately.  One small change could change all the random numbers including the initial values of the network I assume.)

graph2tac/loader/data_server.py
mirefek: I have to do the inverse, and
```
        graphid_to_fname = {
            file_data.graph : str(fname)
            for fname, file_data in data_server._data.items()
        }
```
is in my view the most intuitive way of writing an inverse. Yes, I could keep it in a dictionary, and not convert to a list (but lists have some further advantages).

graph2tac/loader/data_server.py
```diff
@@ -171,21 +172,95 @@ def cluster_to_graph(self, cluster):
...
+class SplitByHash(Splitter):
+    def __init__(self, proportions : list[int], random_seed : int):
+        self.proportions = proportions
+        self.random_seed = random_seed
+    def lemma(self, d : Definition) -> int:
+        # to make it identical to vasily's loader
+        ident_64 = np.array(d.node.identity, dtype = np.uint64).item()
+        return get_split_label(ident_64, self.proportions, self.random_seed)
+    def definition_cluster(self, d : list[Definition]) -> int:
+        return 0
```
jasonrute, line 196: Can we change this 0 to TRAIN?

graph2tac/loader/data_server.py
```diff
@@ -171,21 +172,95 @@ def cluster_to_graph(self, cluster):
...
+        graphid_to_fname = [
+            graphid_to_fname[i]
+            for i in range(len(data_server._data))
+        ]
+
+        # conversion from graph indices to labels + check that all prefixes are used
+        unused_prefixes = set().union(*[prefixes for label,prefixes in self.prefixes_per_label])
+        graphid_to_label = []
+        for fname in graphid_to_fname:
+            label = 0
```
jasonrute, line 225: Can we also change this 0 to TRAIN?

mirefek:
Ah, makes sense, Fidel was not paying attention to emulating Vasily's split (which is used in the hash split).

graph2tac/loader/data_server.py
mirefek: done

graph2tac/loader/data_server.py
LasseBlaauwbroek: Okay, this feels a little convoluted, but I see your point.

graph2tac/loader/data_server.py
jasonrute: Sorry, can you also look at all the other `0`'s and `1`'s in this data_server.py and train.py and see if they should be TRAIN or VALID?  Sorry to be so pedantic.

mirefek: I think I did in the last commit, or have I missed some other?

jasonrute (approved):
LGTM.  Besides some of my pedantic comments, once we get the tests working, I think it should be good to merge.

graph2tac/tfgnn/dataset.py
```diff
@@ -172,25 +158,23 @@ def graph_constants(self) -> GraphConstants:
...
 
-            train_proofstates, valid_proofstates = self.proofstates(split=split,
-                                                                    split_random_seed=split_random_seed,
-                                                                    shuffle=False)
+            train_proofstates, valid_proofstates = self.proofstates(shuffle=False)
 
             proofstate_datasets = {'train_proofstates': train_proofstates.batch(self.STATISTICS_BATCH_SIZE),
                                    'valid_proofstates': valid_proofstates.batch(self.STATISTICS_BATCH_SIZE)}
-            definition_dataset = self.definitions().batch(self.STATISTICS_BATCH_SIZE)
+            definition_dataset = self.definitions(0).batch(self.STATISTICS_BATCH_SIZE) # TODO: what are we using definitions for here?
```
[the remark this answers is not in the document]
jasonrute: This 0 should be TRAIN

graph2tac/tfgnn/dataset.py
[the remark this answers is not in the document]
jasonrute: This 0 should be TRAIN

graph2tac/tfgnn/train.py
```diff
@@ -310,11 +306,9 @@ def run(self,
                                  metrics=self.prediction_task.metrics())
 
         # get training data
-        train_proofstates, valid_proofstates = self.dataset.proofstates(split=split,
-                                                                        split_random_seed=split_random_seed,
-                                                                        shuffle=True)
+        train_proofstates, valid_proofstates = self.dataset.proofstates(shuffle=True)
         if self.definition_task:
-            definitions = self.dataset.definitions(shuffle=True)
+            definitions = self.dataset.definitions(0, shuffle=False)
```
jasonrute, line 311: This 0 should be TRAIN

mirefek:
I updated the tests but since I checked them on my machine, they will likely not work for others...

LasseBlaauwbroek:
The tests pass for me :-)

mirefek:
Fine, I merge, further little bugs will get fixed later.

LasseBlaauwbroek:
Thanks Mirek!

PR #102, opened 2023-02-03
Further cleanup of unused things

LasseBlaauwbroek:
All of this appears to either be unused or out of date. (Someone please press the button for me when they are satisfied with this change.)


jasonrute:
Two questions:
* Did you verify that this code isn't used anywhere anymore?  For example, I thought that `Datapoint` was used, but maybe it is now removed from everywhere.
* The point of Datapoint was as a way to inspect the data.  I think it is important to be able to inspect the data at multiple points in the pipeline.  We never know if the pipeline is screwing something up and our data is coming out a mess.  Or maybe we just want to get statistics on the data.  If Datapoint has already been removed we can deal with it later, but just something for us to think about.

LasseBlaauwbroek:
Yes, a grep of `Datapoint` shows only one result (the definition itself). Note that I'm not at all against having debugging data, but I suggest that we do it in the way we discussed yesterday.

PR #103, opened 2023-02-03
Add example training commands to README.

LasseBlaauwbroek:
(mostly for my own benefit, so I don't have to recreate this thing.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,13 @@ The python `graph2tac` package can be installed with `pip install` in a standard
 - See `g2t-train-tfgnn --help` to train the TF-GNN model
 - See `g2t-server --help` to launch the python server for evaluation
 - See `g2t-tfgnn-predict-graphs --help` to evaluate prediction tfgnn networks on supervised data, and plot results
+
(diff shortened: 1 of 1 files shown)


jasonrute:
In the past, we would never remember to update this stuff if we change things.  So we decided to rely more on `--help`.  But if we will keep this up-to-date, then this is good and helpful.

LasseBlaauwbroek:
Yeah, the other alternative for me is for all the yaml files to be either
1. Automatically selected when nothing is specified on cmd-line
2. One big file that is composed of sub-yamls so that you don't have to specify 5+ files.
3. Make sure the yaml file names correspond exactly to the cmd-line name. Otherwise I'm having to guess which file goes where.

LasseBlaauwbroek:
(I generally don't understand the design of this. I would think that either everything is specified on cmd-line, or everything is specified in a config file. But now, you have to specify multiple config files on cmd-line, that in turn contain multiple options. This doesn't make sense to me.)

jasonrute:
Ok, let's go with this for now.  (But I don't disagree with you.)

PR #105, opened 2023-02-04
Reverse order of recalculated definitions

mirefek:
Signed-off-by: Mirek Olsak <mirek@olsak.net>


jasonrute:
Let's use #106 instead.

PR #106, opened 2023-02-05
Fix bugs in predict server

jasonrute:
* Fix bug where definitions are updated in the reverse order.
* Added an online check that the order is correct since otherwise it would fail silently.
* Added better type annotations.  (I may have went overboard, but it did help me understand the code and catch some bugs.)
* Fixed a number of small bugs my editor found.  Mostly these were in unusual paths through the code that are not traveled during typical running of the server, but for which the code would likely crash if it went down this path.

### DIFF
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -31,8 +31,9 @@ def apply_temperature(confidences: ArrayLike, temperature: float) -> NDArray[np.
     assuming the truncated tail of the probability distribution
     is clustered in one unseen element
     """
-    res = np.array(confidences, np.float64)**(1/temperature)
-    res /= res.sum() + (1-sum(confidences))**(1/temperature)
+    confidences = np.array(confidences, np.float64)
+    res = confidences**(1/temperature)
+    res /= res.sum() + (1-confidences.sum())**(1/temperature)
     return res
 
 @dataclass
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -330,12 +330,12 @@ class PredictServer(AbstractDataServer):
             if tactic.ident not in self._tactic_to_i
         ]
         unaligned_definitions = [
-            d for d in msg.definitions.definitions
+            (d.node.nodeid, d) for d in msg.definitions.definitions
             if d.name not in self._def_name_to_i
         ]
 
         unaligned_tactics = sorted(unaligned_tactics)
-        unaligned_definitions = sorted(unaligned_definitions)
+        unaligned_definitions = [d for _, d in sorted(unaligned_definitions)]
         return CheckAlignmentResponse(
             unknown_tactics = unaligned_tactics,
             unknown_definitions = unaligned_definitions,
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -1,9 +1,12 @@
 import argparse
 from contextlib import contextmanager
+from dataclasses import dataclass
 import sys
 import socket
 from pathlib import Path
+from typing import IO, Optional
 import numpy as np
+from numpy.typing import NDArray, ArrayLike
 import tqdm
 import os
 import uuid
@@ -15,12 +18,14 @@ import yaml
 
 from pytact.data_reader import (capnp_message_generator,
                                 TacticPredictionGraph, TacticPredictionsGraph,
-                                GlobalContextMessage, CheckAlignmentMessage, CheckAlignmentResponse)
+                                GlobalContextMessage, CheckAlignmentMessage, CheckAlignmentResponse,
+                                ProofState, OnlineDefinitionsReader)
 from graph2tac.common import logger
-from graph2tac.loader.data_classes import *
-from graph2tac.loader.data_server import AbstractDataServer, DataServer
+from graph2tac.loader.data_classes import LoaderProofstate, ProofstateContext, ProofstateMetadata
+from graph2tac.loader.data_server import AbstractDataServer
+from graph2tac.predict import Predict
 
-def apply_temperature(confidences, temperature):
+def apply_temperature(confidences: ArrayLike, temperature: float) -> NDArray[np.float64]:
     """
     Apply the temperature,
     assuming the truncated tail of the probability distribution
@@ -57,7 +62,7 @@ class LoggingCounters:
     """Max time (in seconds) spent on single predict since last 'initialize'"""
     argmax_t_predict: int = -1
     """Index of last maximum time single predict since last 'initialize'"""
-    t_coq: int = 0
+    t_coq: float = 0.0
     """Total time (in seconds) spent waiting on Coq since last 'initialize'"""
     n_coq: int = 0
     """Number of 'prediction' requests made by Coq"""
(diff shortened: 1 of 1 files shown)


graph2tac/loader/predict_server.py
```diff
@@ -220,8 +230,21 @@ def _enter_coq_context(self, definitions, tactics):
...
+                for n in used_nodes:
+                    assert n < prev_defined_nodes or n in defined_nodes, (
+                        f"Definition clusters out of order. "
+                        f"Attempting to compute definition embedding for node labels {new_defined_nodes} "
+                        f"({cluster_state.definition_names}) without first computing "
+                        f"the definition embedding for node label {n} used in that definition."
+                    )   
+                for n in new_defined_nodes:
+                    defined_nodes.add(n)
+
```
jasonrute, line 247: This is probably the most controversial thing.  I'm checking every time that the definitions are in the correct order.  My thinking is that if we ever mess this up it could be *weeks or months* before we notice.  Also, it is difficult to make a unit test for this, since it relies on a contract with coq-tactician-reinforce.  (Also, the added time to the code is negligible.)

graph2tac/loader/predict_server.py
```diff
@@ -200,18 +205,23 @@ def _enter_coq_context(self, definitions, tactics):
...
             def_clusters_for_update = list(definitions.clustered_definitions)
+            def_clusters_for_update.reverse()
+            prev_defined_nodes = self._base_node_label_num
             logger.info(f"Prepared for update all {len(def_clusters_for_update)} definition clusters")
         elif self.config.update_new_definitions:
             def_clusters_for_update = [
                 cluster for cluster in definitions.clustered_definitions
                 if self._def_node_to_i[cluster[0].node] >= self._num_train_nodes
             ]
+            def_clusters_for_update.reverse()
```
LasseBlaauwbroek, line 219: I would pull the `reverse`  out of the if-then else.

graph2tac/loader/predict_server.py
```diff
@@ -319,15 +342,18 @@ def check_alignment(msg):
             unalignedDefinitions = unaligned_definitions,
         )
 
-def prediction_loop(predict_server, capnp_socket, record_file):
-    if predict_server.config.with_meter:
-        capnp_socket = tqdm.tqdm(capnp_socket)
+def prediction_loop(predict_server: PredictServer, capnp_socket: socket.socket, record_file: Optional[IO]):
```
LasseBlaauwbroek, line 345: Shouldn't `IO` be `BinaryIO`?

graph2tac/loader/predict_server.py
LasseBlaauwbroek: The more checks the better! (Unless it causes unacceptable slowdown.)

graph2tac/loader/predict_server.py
```diff
@@ -319,15 +342,18 @@ def check_alignment(msg):
...
+    if predict_server.config.with_meter:
+        message_generator_iterator = tqdm.tqdm(message_generator)
+    else:
+        message_generator_iterator = message_generator
     log_cnts = predict_server.log_cnts
     log_cnts.session_idx += 1
     log_cnts.thm_idx = -1
 
-    for msg in message_generator:
+    for msg in message_generator_iterator:
```
LasseBlaauwbroek, line 356: That's a very strange name. Why both `generator` and `iterator`?

LasseBlaauwbroek (approved):
Some minor comments, otherwise looks good.

graph2tac/loader/predict_server.py
jasonrute: Good point.  I was unsure how strong of a contract `capnp_message_generator` has on record_file, but I should have just looked at the code.  Fixed.

LasseBlaauwbroek: As an aside: Is your editor able to retrieve the type annotations and documentation from `data_reader.pyx`? Mine isn't, which is extremely annoying...

graph2tac/loader/predict_server.py
mirefek: You could also add a check that the new defined nodes are above `prev_defined_nodes`, and were not defined before.

graph2tac/loader/predict_server.py
```diff
@@ -319,15 +342,18 @@ def check_alignment(msg):
...
-def prediction_loop(predict_server, capnp_socket, record_file):
-    if predict_server.config.with_meter:
-        capnp_socket = tqdm.tqdm(capnp_socket)
+def prediction_loop(predict_server: PredictServer, capnp_socket: socket.socket, record_file: Optional[IO]):
+    
     message_generator = capnp_message_generator(capnp_socket, record_file)
+    if predict_server.config.with_meter:
+        message_generator_iterator = tqdm.tqdm(message_generator)
+    else:
+        message_generator_iterator = message_generator
```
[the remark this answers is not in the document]
mirefek: Naming it `message_iterator` instead of `message_generator_iterator` is an improvement but I still don't see the point of your change in the first place -- I find it quite redundant, it is just a little decoration of the generator, and there is no meaning in using the undecorated version afterwards

graph2tac/loader/predict_server.py
jasonrute: Changed the name to just `message_iterator`.  By the way, here is what is going on.
```
message_generator = capnp_message_generator(capnp_socket, record_file)
```
returns a generator.  Generators have a method `send` which we use for checking alignment.  If we wrap the generator in `tqdm` then it no longer is a generator and we can't use `send`. So I need to keep around the inner generator as well as the wrapped iterator if we want to keep this `tqdm` functionality.

graph2tac/loader/predict_server.py
jasonrute: No.  But I think you can add type stubs to make this information available.  I was going to approach you about it.

graph2tac/loader/predict_server.py
mirefek: Oh, I see

LasseBlaauwbroek: I see. It seems kind of silly for `tqdm` not to support generators. But whatever.

mirefek (approved):
As I mentioned, you could also add a few checks for `new_defined_nodes` but in general I these changes make sense.

graph2tac/loader/predict_server.py
```diff
@@ -302,32 +324,33 @@ def predict(self, proof_state):
...
             if tactic.ident not in self._tactic_to_i
         ]
         unaligned_definitions = [
-            d.node.nodeid for d in msg.definitions.definitions
+            d for d in msg.definitions.definitions
             if d.name not in self._def_name_to_i
         ]
 
-        unaligned_tactics = sorted(unaligned_tactics)
-        unaligned_definitions = sorted(unaligned_definitions)
```
jasonrute, line 316: Lasse said sorting isn't necessary here, so I removed it.

graph2tac/loader/predict_server.py
jasonrute: I could be mistaken, but I think if we only check the new nodes, then that doesn't cover a case where a used node is somehow a bad index which is too high.  One way this could come up is if we end up repeating a node label in the same graph which defines it.  Or if somehow we end up with two definitions which mutually depend on each other.  (which shouldn't happen with our clustering).  Or if we don't get the boundary of previous nodes correct and end up skipping a bunch of new nodes leaving them undefined.  So I like this a bit better.  (But yes, to just check the definitions are in order, your approach would be sufficient.)

jasonrute: Sorry, you are suggesting additional checks.  Yes, that makes sense.  I'll add them.

jasonrute: Added.

PR #107, opened 2023-02-08
Mirek/align by ident

mirefek:
For prediction, aligning definitions by node identities instead of names

### DIFF
--- a/graph2tac/loader/data_classes.py
+++ b/graph2tac/loader/data_classes.py
@@ -126,5 +126,6 @@ class GraphConstants:
     tactic_index_to_hash: list[int]
     global_context: list[int]
     label_to_names: list[str]
+    label_to_ident: list[int]
     label_in_spine: list[bool]
     max_subgraph_size: int
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -156,13 +156,14 @@ class PredictServer(AbstractDataServer):
         }
 
         self._node_i_to_name = model.get_label_to_name()
+        self._node_i_to_ident = model.get_label_to_ident()
         self._node_i_in_spine = model.get_label_in_spine()
         self._num_train_nodes = len(self._node_i_to_name)
 
         del self._def_node_to_i # not usable without a coq_context
-        self._def_name_to_i = {
+        self._def_ident_to_i = {
             name : i
-            for i,name in enumerate(self._node_i_to_name)
+            for i,name in enumerate(self._node_i_to_ident)
             if i >= self._base_node_label_num and self._node_i_in_spine[i]
         }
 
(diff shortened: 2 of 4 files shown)


graph2tac/loader/predict_server.py
```diff
@@ -156,13 +156,14 @@ def __init__(self,
...
         self._node_i_in_spine = model.get_label_in_spine()
         self._num_train_nodes = len(self._node_i_to_name)
 
         del self._def_node_to_i # not usable without a coq_context
-        self._def_name_to_i = {
+        self._def_ident_to_i = {
             name : i
-            for i,name in enumerate(self._node_i_to_name)
+            for i,name in enumerate(self._node_i_to_ident)
             if i >= self._base_node_label_num and self._node_i_in_spine[i]
```
LasseBlaauwbroek, line 167: I would expect `self._node_i_in_spine[i]` to go away (otherwise there is not much improvement here). You'll still have to deal with some collisions though.

LasseBlaauwbroek:
I wrote a quick script to investigate definitions with duplicate identities:
```python
from pathlib import Path
import sys
from pytact.data_reader import data_reader
from collections import defaultdict

identities = defaultdict(list)

dataset_path = Path(sys.argv[1]).resolve()
with data_reader(dataset_path) as data:
    for datafile in data.values():
        for d in datafile.definitions():
            identities[d.node.identity].append(f"{datafile.filename} : {d.name}")
for i, ls in identities.items():
    if len(ls) > 1:
        print(f"{i}")
        for d in ls:
            print(f"    {d}")
```

This shows that duplicates are caused by three categories:
1. Section variables that are repeated amongst multiple sections (these are the definitions with a fully unqualified name like `A`).
2. Definitions inside of module functors that are being repeated (these are definitions with names like `M.min_l`).
3. Some outliers that are related to the `abstract` tactic (we can ignore this I think).

LasseBlaauwbroek:
Repeating a message I sent to Mirek in slack:

> When you align with any nodes that are not in the spine, you'll have to come up with a way to deal with collisions. I can see a couple of methods, not sure which one is the best:
> 1. When a collision occurs, just align to a random definition in that collision.
> 2. When a collision occurs, don't align it at all.
> 3. During training, you can already resolve any collision by lumping all definitions with the same identity together. Not sure what the effect of that would be.
> 4. You can try to infer the 'correct' definition to align to by trying to find a 'consistent' set of definitions in the training data (that is, definitions that all occur in the same sub-spine) such that the most definitions are aligned. This is the most complicated solution, but probably also the most correct solution.

jasonrute:
When Lasse is happy with this, I am.  I don't think I understand the fine points anyway.

graph2tac/loader/data_classes.py
```diff
@@ -126,5 +126,6 @@ class GraphConstants:
     tactic_index_to_hash: list[int]
     global_context: list[int]
     label_to_names: list[str]
+    label_to_ident: list[int]
     label_in_spine: list[bool]
```
LasseBlaauwbroek, line 130: Do we still need `label_in_spine`? And what about `label_to_names`?

LasseBlaauwbroek:
I think that the correct choice also very much depends on the semantics of the model. Not sure if I'm qualified to determine that. Maybe we can discuss it tomorrow.

PR #108, opened 2023-02-08
Remove TF2 model

jasonrute:
This removes the TF2 model.  To access the TF2 model at its last working commit, use the tagged release [`last-working-tf2`](https://github.com/IBM/graph2tac/tree/last-working-tf2).

I removed the `tf2` directory, all the old tf2 testing code, any tf2 code in the predict server, the `g2t-train` entry point, and a number of python dependencies which are no longer needed.

I checked that the tests still work, and that the g2t-server still works (on the tfgnn model).   Nonetheless, it is possible I deleted some transitive dependency that is used in some obscure part of the code, so if we notice that, we can add it back into `setup.py`.  I didn't touch the notebooks, since I think we have decided they are not worth keeping up-to-date.


graph2tac/loader/predict_server.py
```diff
@@ -420,8 +420,8 @@ def parse_args() -> argparse.Namespace:
                         help='checkpoint directory of the model')
 
     parser.add_argument('--arch', type=str,
-                        default='tf2',
-                        help='the model architecture tfgnn or tf2 or hmodel (current default is tf2)')
+                        default='tfgnn',
```
LasseBlaauwbroek, line 423:
```suggestion
                        default='tfgnn',
                        choices=['tfgnn', 'hmodel'],
```
While you're at it, maybe make the `--help` output a little nicer?

LasseBlaauwbroek (approved):
Excellent. One suggestion for a drive-by improvement. Otherwise looks good.

LasseBlaauwbroek:
Perhaps it's a good idea to tag the `tf` branch, so that we can always find ourselves back to a working state of it?

graph2tac/loader/predict_server.py
jasonrute: What else do you have in mind for the output of help?

jasonrute:
> Perhaps it's a good idea to tag the `tf` branch, so that we can always find ourselves back to a working state of it?

Should we call the tag `tf2`, or do you have a suggested naming scheme?

graph2tac/loader/predict_server.py
jasonrute: Or were you just referring to that particular change?

LasseBlaauwbroek:
I'd say something like `last-working-tf2`. Whatever. And perhaps also document somewhere (perhaps in the tag commit message) what the known working commit of `coq-tactician-reinforce` is.

graph2tac/loader/predict_server.py
LasseBlaauwbroek: Well, this was just meant as a simple drive-by improvement. But here is a list of potential further improvements:
1. `--port` and `--host` are only used when `--tcp` is active. I'd specify this using something like `--tcp 0.0.0.0:3333`.
2. `--update-new-definitions` and `--update-all-definitions` are mutually exclusive. You can encode this with something like ` update_group = parser.add_mutually_exclusive_group()`.
3. `--log-level` and `--tf_log_level` should also have a `choices` entry.
4. Do we really need to differentiate `--with-meter` and `--progress-bar`?

LasseBlaauwbroek: Also, `--model` is specified as optional, but I don't think it is.

LasseBlaauwbroek:
Also, the README is in need of updating.

PR #109, opened 2023-02-22
Mirek/simplify2

mirefek:
This is still a work in progress -- the ultimate goal is to minimize the conversion between LoaderGraph and GraphTensor, put everything into one class (DataServer), and get rid of dataset.py. On the other hand, we thought with Lasse that it might be a good idea to already merge this, so that our git repo doesn't get too branched.


graph2tac/loader/data_classes.py
```diff
@@ -1,118 +1,79 @@
...
-    """String repr for a dataclass, but removing the array data."""
-    kws = {k: f"array(shape={v.shape}, dtype={v.dtype})" if isinstance(v, np.ndarray) else repr(v) for k,v in obj.__dict__.items()}
-    return obj.__class__.__qualname__ + "(" + ", ".join(k + "=" + v for k,v in kws.items()) + ")"
-
-
-@dataclass(repr=False)
-class LoaderGraph:
-    """A single graph as it comes from the loader"""
-    nodes: NDArray[np.uint32]  # [nodes]
-    """Node class labels.  Shape: [number of nodes]"""
```
LasseBlaauwbroek, line 15: Maybe we can keep the docs, even if they are not official docstrings?

LasseBlaauwbroek (approved):
Looks good to me, assuming that everything still works :-)

The only question is if we find the training slowdown to be acceptable.

graph2tac/loader/data_classes.py
jasonrute: Also, why can't we just do what Lasse had originally suggested with the named tuples.  Then we can have both namedtuples and good editor hints including doc strings.  Is there something that makes that approach not work?

jasonrute:
```
from typing import NamedTuple

class Card(NamedTuple):
    """This is a card type."""
    suit: str
    rank: str
```

jasonrute: Or is it that this has too much boilerplate since it requires writing code for two classes instead of one?

mirefek: I had some trouble with it with my first experiments but I admit it probably would work. On the other hand, we need the tensorflow typing anyway, and I prioritize having nice way of managing tensorflow typings over editor features. This is mainly a helper datastructure between pytact, and the GraphTensor, and when we decide to put another information to the network, we want to be able to change the code at as little places as possible. Up to discussion (tomorrow?)

mirefek: If we could do both with some smart decorator, it could be cool. Does `numpy.typing` allow shape hints like tensorflow?

graph2tac/tfgnn/train.py
```diff
@@ -306,9 +306,10 @@ def run(self,
                                  metrics=self.prediction_task.metrics())
 
         # get training data
-        train_proofstates, valid_proofstates = self.dataset.proofstates(shuffle=True)
+        train_proofstates = self.dataset.proofstates(TRAIN, True)
+        valid_proofstates = self.dataset.proofstates(VALID, False)
         if self.definition_task:
-            definitions = self.dataset.definitions(TRAIN, shuffle=False)
+            definitions = self.dataset.definitions(TRAIN, False)
```
jasonrute, line 312: This is actually a bug not from this PR but from the previous one.  I discovered it this week and it leads to a fairly large regression in our model.  Can you make this `shuffle=True` again?

jasonrute: Definitions need to be shuffled like all other training data.  (Also, there is a good reason to shuffle definitions for validation as well.  Since there are more definitions than validation proof states (I think), to get a good validation statistic for definition loss, we should have a random subset of definitions.  Of course, one day we may just want separate definitions for validation and training.)

mirefek: done

jasonrute:
I'm still looking over the code, but I just want to check that you tested that the predict server and hmodel still run.  This code effects both.

graph2tac/tfgnn/predict.py
```diff
@@ -333,23 +332,17 @@ def compute_new_definitions(self, new_cluster_subgraphs: List[LoaderDefinition])
             raise RuntimeError('cannot update definitions when a definition task is not present')
 
         assert len(new_cluster_subgraphs) == 1
-        new_cluster_subgraph = DataServerDataset._loader_to_definition_data(new_cluster_subgraphs[0])
+        new_cluster_subgraph = self._dataset._loader_to_definition_graph_tensor(new_cluster_subgraphs[0])
```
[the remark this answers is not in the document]
jasonrute: Another quick option is just to compile this function.

graph2tac/tfgnn/dataset.py
```diff
@@ -516,6 +196,11 @@ def _make_bare_graph_tensor(node_labels: tf.Tensor,
         @param edge_labels: tf.Tensor of edge labels (dtype=tf.int64)
         @return: a GraphTensor object that is compatible with the `bare_graph_spec` in `graph_schema.py`
         """
+        node_labels = tf.convert_to_tensor(graph.nodes, dtype = tf.int64)
+        edge_labels = tf.convert_to_tensor(graph.edge_labels, dtype = tf.int64)
+        sources = tf.convert_to_tensor(graph.edges[:,0], dtype = tf.int32)
+        targets = tf.convert_to_tensor(graph.edges[:,1], dtype = tf.int32)
+
```
jasonrute, line 203: We shouldn't need these conversions if this is compiled and the TF types are specified in the spec.

jasonrute: And elsewhere in this file.

graph2tac/tfgnn/dataset.py
```diff
@@ -585,89 +255,21 @@ def _make_proofstate_graph_tensor(cls,
...
-        @param loader_graph: Tuple of tf.Tensor objects encoding the graph in the loader's format
-        @param num_definitions: tf.Tensor for the number of labels being defined (dtype=tf.int64)
-        @return: a GraphTensor object that is compatible with the `definition_graph_spec` in `graph_schema.py`
-        """
-        node_labels, edges, edge_labels, _ = loader_graph
-        sources = edges[:, 0]
-        targets = edges[:, 1]
-
-        bare_graph_tensor = cls._make_bare_graph_tensor(node_labels, sources, targets, edge_labels)
+    def _loader_to_definition_graph_tensor(self, defn: LoaderDefinition) -> tuple:
```
jasonrute, line 265: Fix output type.

jasonrute (commented):
It mostly looks good but here are some comments.  Some are bugs.  Others are code cleanup stuff.  Others are just comments.  I would say let's fix the clear bugs and the code stuff that isn't going to be totally rewritten (since otherwise we will both forget about this).  The stuff that is going to be rewritten is less of a concern.

Also, I ran you model and I'm seeing a regression on the local argument accuracy/loss which I'm looking into and I contacted you about.

Finally, I'd also like to run the trained model through a short benchmark just to get a sense of timings/memory changes and also to see if there are any other issues.

jasonrute:
I benchmarked the model.  I don't see any significant memory leaks, but I just want to point out that computing the definitions now takes twice as long likely because the `loader_to_..._graph_tensor` functions are not compiled.  This is probably even more important for definitions than proof states since `loader_to_definition_graph_tensor` function is doing the string vectorization inside that function.  Since we are changing that part code, this it isn't a big deal right now, but just wanted to make everyone aware of it.

Previous model:
```
graph2tac:SUMMARY - Thm Summary:
[g2t-sum] UUID	88f6bcf8-a7f7-11ed-b8e9-b03af2b6059f
[g2t-sum] Session	0
[g2t-sum] Theorem	1
[g2t-sum] Annotation	File "./affine_classiques.v", line 89, characters 0-43:
[g2t-sum] Initialize|Network build time (s)	0.145384
[g2t-sum] Initialize|Def clusters to update	10976
[g2t-sum] Initialize|Def update time (s)	75.101830
[g2t-sum] Predict|Messages cnt	779
[g2t-sum] Predict|Avg predict time (s/msg)	0.263283
[g2t-sum] Predict|Max predict time (s/msg)	2.375712
[g2t-sum] Predict|Max time predict msg ix	0
[g2t-sum] Predict|Avg Coq wait time (s/msg)	106.254888
[g2t-sum] Memory|Total memory (MB)	7521.231
[g2t-sum] Memory|Physical memory (MB)	1606.197
[g2t-sum] Memory|Total mem diff (MB)	98.324
[g2t-sum] Memory|Physical mem diff (MB)	161.767
[g2t-sum] Avg CPU (%)	153.0
```
Current model
```
graph2tac:SUMMARY - Thm Summary:
[g2t-sum] UUID	71ce8a1a-b3d7-11ed-b083-00001049fe80
[g2t-sum] Session	0
[g2t-sum] Theorem	1
[g2t-sum] Annotation	File "./affine_classiques.v", line 89, characters 0-43:
[g2t-sum] Initialize|Network build time (s)	0.150906
[g2t-sum] Initialize|Def clusters to update	10976
[g2t-sum] Initialize|Def update time (s)	202.253616
[g2t-sum] Predict|Messages cnt	289
[g2t-sum] Predict|Avg predict time (s/msg)	0.278455
[g2t-sum] Predict|Max predict time (s/msg)	2.376219
[g2t-sum] Predict|Max time predict msg ix	0
[g2t-sum] Predict|Avg Coq wait time (s/msg)	41.824722
[g2t-sum] Memory|Total memory (MB)	7269.532
[g2t-sum] Memory|Physical memory (MB)	1289.994
[g2t-sum] Memory|Total mem diff (MB)	494.764
[g2t-sum] Memory|Physical mem diff (MB)	114.852
[g2t-sum] Avg CPU (%)	166.9
```

mirefek:
Thanks for the info, that should be easy to fix. Have you figured out what is causing the bug (?) with local arguments?

mirefek:
This commit should fix the definition recalculation speed

LasseBlaauwbroek:
W.r.t. the issue with tensorflow's excessive logging: To fix this, wouldn't it be enough to modify `predict_server.py` to delay the import of
```
from dataclasses import dataclass
from graph2tac.predict import Predict
```
until after
```
    os.environ['TF_CPP_MIN_LOG_LEVEL'] = tf_env_log_levels[config.tf_log_level]
```
on line 518?

jasonrute:
@LasseBlaauwbroek You would also have to move `from graph2tac.loader.data_server import AbstractDataServer` which is used pretty early, right?.  I think see three feasible options:
1. Except that our whole codebase depends on TF and just move `os.environ['TF_CPP_MIN_LOG_LEVEL'] = tf_env_log_levels[config.tf_log_level]` to the top of the file before any of the `import graph2tac ...` lines.  This is what we did for TF2 training.  The only big downside is that argument parsing has to wait for TF to load, so `g2t-server --help` will take a long time to run.
2. Make another entry point (in another file) for the server which does arg parsing, loaders the rest of the server code, and runs it.
3. Rearrange our code a little bit.  I see a few possibilities:
  * Try to move tensorflow out of a few files or break up files to do so.  The main challenge is that our datatypes now depend on tensorflow, so we either get rid of them, that dependency.
  * Use dependency injection to make an even more abstract `AbstractDataServer` and pass it to `PredictServer` as a parameter.  (I don't know however if this will work, since even the signature of the AbstractDataServer methods probably use our TF-dependent datatypes now.  But, honestly, even if we kept AbstractDataServer the same, I'd still like this approach better since I think the PredictServer is not a DataServer and calls like `self._downward_closure([root])` are confusing`.  I'd rather see something like `self.data_server.get_graph(root)`.  Dependency injection also makes it possible to do unit testing, since you can swap out the real DataServer with a fake version.

But honestly, I think we will just end up with the first (or maybe the second) option since it is easiest, and really any real world use of the server will require loading tensorflow.

LasseBlaauwbroek:
Option (1) is completely fine with me.

LasseBlaauwbroek:
But note that `config.tf_log_level` comes from the parser. So that means that the parser has to be created earlier as well (and will be fast, unlike what you say). All of that is fine with me.

jasonrute:
I don't know how to create the argparser earlier but still have it run from `main` (which we need to use entry points) unless we go with option 2 (a file for the entry point `main` function with an arg parser, and another file which builds all the classes for the server).

LasseBlaauwbroek:
Option (2) is fine with me as well. Or just move all of the imports into the `main` function. Whatever.

mirefek:
Do you want to do this already in this PR?

LasseBlaauwbroek:
Up to Jason.

jasonrute:
How about we do this right now.  Just add `os.environ['TF_CPP_MIN_LOG_LEVEL'] = tf_env_log_levels[config.tf_log_level]` before any of the graph2tac imports.  This should turn off tensorflow warnings until we get to another point in the code where we turn them back on.  Then we will make a point to split up the this file when you are down refactoring it Mirek.  (The main useful warnings about about not having GPUs, but I don't run benchmarking with GPUs anyway.)

graph2tac/tfgnn/dataset.py
```diff
@@ -5,56 +5,98 @@
...
+    definition_names_spec = tf.TensorSpec(shape=(None,), dtype=tf.string, name='definition_names')
+
+    proofstate_data_spec = (state_spec, action_spec, graph_id_spec)
+    definition_data_spec = (loader_graph_spec, num_definitions_spec, definition_names_spec)
 
     def __init__(self,
-                 graph_constants: GraphConstants,
+                 data_dir: Path,
+                 split_method : str,
+                 split,
```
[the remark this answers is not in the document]
mirefek: yes, it is a plan

graph2tac/tfgnn/predict.py
```diff
@@ -211,10 +212,10 @@ def __init__(self,
 
         dataset_yaml_filepath = log_dir / 'config' / 'dataset.yaml'
         with dataset_yaml_filepath.open('r') as yml_file:
-            dataset = Dataset(graph_constants=graph_constants, **yaml.load(yml_file, Loader=yaml.SafeLoader))
+            dataset = DataServerDataset(None, None, None, graph_constants=graph_constants, **yaml.load(yml_file, Loader=yaml.SafeLoader))
```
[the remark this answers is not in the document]
mirefek: I am going to change this next time

graph2tac/tfgnn/dataset.py
mirefek: on the other hand, I think it is good to have a code which works even outside tensorflow. Otherwise, I was getting very incomprehensible error messages.

PR #111, opened 2023-02-27
Upgrade pytact to fix Cython version

LasseBlaauwbroek:


jasonrute:
I notice this commit is five commits ahead of the previous coq-tactician-reinforce commit.   Was there any change to the`pytact` code in those commits?  If so, did you run the tests?  (But other than that, I'm happy with this as a temporary fix until things are resolved.)

PR #112, opened 2023-03-07
Upgrade to v15

LasseBlaauwbroek:
This PR upgrades graph2tac to v15. This is preliminary, because I'm still generating the final dataset.

This PR is split into well-defined commits. I recommend reviewing commit-by-commit.

I also updated `predict_server.py` to work with the new incremental protocol. It seems to be functioning, but the code is a bit awkward because the model currently doesn't support incremental updating. I've left a FIXME in the code, for someone to remedy this.

One issue that I've not investigated: When running the prediction server with `--update-all-definition` I'm getting:
```
RuntimeError: cannot update definitions when a definition task is not present
```
That seems to indicate that I've trained the model with the wrong task? I used the default parameters like this:
```
g2t-train-tfgnn --data-dir tests/data/mini_stdlib/dataset/  --dataset-config graph2tac/tfgnn/default_dataset_config.yml --prediction-task-config graph2tac/tfgnn/default_global_argument_prediction.yml --trainer-config graph2tac/tfgnn/default_trainer_config.yml --run-config graph2tac/tfgnn/default_run_config.yml --log model/
```
What is going on here?


jasonrute:
(On my phone so didn't double check all this.)  I think you are missing the --definition_task when training.  Therefore it trains without a definition task.  If you leave off --update-all-definitions when running the predict server, this should all work with the model you trained.

jasonrute:
Also for simple testing, you can get an already trained model in the temporary directory where the test results are stored after you run `pytest`.  I think the test outputs tell you were that is, but if not just look through your `/tmp` directory.

LasseBlaauwbroek:
Confirmed.

As an aside: It feels to me that the predict server has become significantly slower than it used to be. (No hard numbers, just what I remember.)

LasseBlaauwbroek:
I believe this is ready.

graph2tac/loader/predict_server.py
```diff
@@ -343,29 +343,25 @@ def check_alignment(self, msg: CheckAlignmentMessage) -> CheckAlignmentResponse:
             unknown_definitions = unaligned_definitions,
         )
 
-def prediction_loop(predict_server: PredictServer, capnp_socket: socket.socket, record_file: Optional[BinaryIO]):
-    
-    message_generator = capnp_message_generator(capnp_socket, record_file)
+def prediction_loop(predict_server: PredictServer, context: GlobalContextMessage):
```
jasonrute, line 346: I'm a bit confused by the types here.  This is called below as
```
prediction_loop(predict_server, capnp_message_generator(capnp_socket, record_file))
```
And `capnp_message_generator` returns type
```
Generator[GlobalContextMessage | CheckAlignmentMessage, None | CheckAlignmentResponse, None])
```
I don't even understand how this runs, much alone type checks.  If context is a generator, how does `context.prediction_requests` work?  It obviously does run, so I must be missing something.  How does this work?

jasonrute: Nevermind.  I was looking at the wrong branch of coq-tactician-reinforce.  It no longer returns a generator, but a `GlobalContextMessage` which is exactly what we need to make this work.

PR #116, opened 2023-03-10
Fix bugs

jasonrute:
Fix a bug related to unicode.  Note we still don't tokenize unicode except to make all unicode characters tokenize to `1` uniformly.  But now it doesn't crash when we see a unicode character.

Fix a verbose warning in the numpy code related to casting.  The casting still works the same as before.

The tests pass.

### DIFF
--- a/graph2tac/loader/data_server.py
+++ b/graph2tac/loader/data_server.py
@@ -232,7 +232,7 @@ class SplitByHash(Splitter):
         self.random_seed = random_seed
     def lemma(self, d : Definition) -> int:
         # to make it identical to vasily's loader
-        ident_64 = np.array(d.node.identity, dtype = np.uint64).item()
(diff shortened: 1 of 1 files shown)


mirefek:
The np.uint64 from splitter could be removed completely. It is there only to copy the behavior of Vasily's loader.
Similarly the string types could be possibly just strings instead of numpy specialities. On the other hand, I get that this is just a bugfix, so I think it can be just merged.

mirefek (approved):
The np.uint64 from splitter could be removed completely. It is there only to copy the behavior of Vasily's loader.
Similarly the string types could be possibly just strings instead of numpy specialities. On the other hand, I get that this is just a bugfix, so I think it can be just merged.

jasonrute:
I'll just merge it now since it is needed for getting v15 to work.

PR #117, opened 2023-03-14
Mirek/simplify3

mirefek:
I am finishing the simplification, and planning to focus on other things now. Highlights:

- Diminishing `DataServerDataset` up to a little class `DataToTFGNN`, currently in `loader/data_server.py`
- TFGNN prediction is now only supporting handling single datapoint, and it is time-optimized for that (as discussed "main" / "function").
- `PredictServer` is no longer a subclass of `AbstractDataServer`, instead of it, there is another class `DynamicDataServer` for the data handling part.
- Dataset configuration is now structured into `DataConfig` (about how to handle the graphs), and `DatasetConfig` (about splits, excluding datapoints, ...) `GraphConstants` is containing a `DataConfig` as an object, not its individual items. It also means that any Dataset configuration files we are using would have to be updated.


mirefek:
By the way, all the tests passed, I also checked prediction and hmodel on a simple example to make sure it doesn't crash.

graph2tac/loader/data_classes.py
```diff
@@ -48,12 +48,13 @@ def namedtuple_with_spec(namedtuple_name, **item_spec_d):
...
+    local_args = tf.TensorSpec([None], dtype=tf.int64),
+    global_args = tf.TensorSpec([None], dtype=tf.int64),
     # Tactic arguments (local or global or none).  Using the indices in the `context` field of the proofstate.
     
-    # If `args[i] == [0, l] for l < local_cxt_size`, then the `i`th arg is a local context variable with index `l`.
-    # If `args[i] == [0, local_cxt_size]`, then the `i`th arg is a none argument (can't be represented).
-    # If `args[i] == [1, g]` for `g > 0`, then the `i`th arg is a global definition with index `g`.
+    # If `local_args[i] == l' for 'l >= 0', then the `i`th arg is a local context variable with index `l`.
+    # If `global_args[i] == g` for `g >= 0`, then the `i`th arg is a global definition with index `g`.
+    # Otherwise, the values are set to '-1'
```
jasonrute, line 57: I really like this!  So much easier to follow.

graph2tac/loader/predict_server.py
```diff
@@ -130,90 +130,132 @@ def summary_log_message(self) -> str:
...
-            stop_at_definitions =  model.graph_constants.stop_at_definitions,
-            symmetrization =  model.graph_constants.symmetrization,
-            add_self_edges =  model.graph_constants.add_self_edges
-        )
-        self.model = model
-        self.config = config
-        self.log_cnts = log_cnts
-        self.current_definitions = None # only with a coq context
+class DynamicDataServer(AbstractDataServer):
+    def __init__(self, graph_constants):
```
jasonrute, line 134:
```suggestion
    def __init__(self, graph_constants: GraphConstants):
```

graph2tac/loader/predict_server.py
```diff
@@ -130,90 +130,132 @@ def summary_log_message(self) -> str:
...
+            metadata=dummy_proofstate_info
+        )
+
+class PredictServer:
+    def __init__(self,
+                 model: Predict,
+                 config: argparse.Namespace,
+                 log_cnts: LoggingCounters,
+    ):
+        self.data_server = DynamicDataServer(model.graph_constants)
```
jasonrute, line 215: Nice!

tests/data/mini_stdlib/params/split_fname_prefix/expected.json
jasonrute, line 1: Do you know why the test numbers changed?

tests/integration/test_training_pipeline.py
```diff
@@ -8,11 +8,13 @@
 from unittest.mock import patch
 import tensorflow as tf
 import warnings
+import random
 
 import graph2tac.tfgnn.train
 
 # this helps with determinism
 tf.config.experimental.enable_op_determinism()
+random.seed(42)
```
jasonrute, line 17: We don't need this, right?
```suggestion
```

graph2tac/loader/data_classes.py
mirefek: It is actually the format Fidel was using in the network. I just skipped the intermediate conversion :-)

tests/data/mini_stdlib/params/normal/expected.json
jasonrute, line 1: Why did the test numbers here change significantly?

graph2tac/tfgnn/predict.py
```diff
@@ -425,18 +358,82 @@ def _expand_arguments_logits(cls,
...
+                graph_tensor_stacked = stack_graph_tensors([graph_tensor_single])
+                inference_output = inference_model_bare({
+                    self.prediction_task.PROOFSTATE_GRAPH: graph_tensor_stacked,
+                    self.prediction_task.TACTIC_MASK: tf.expand_dims(tactic_mask, axis=0),
+                })
+                return inference_output
             self._inference_model_cache[tactic_expand_bound] = inference_model
         return self._inference_model_cache[tactic_expand_bound]
 
+    def _tactic_mask_from_allowed_model_tactics(self, allowed_model_tactics: Optional[Iterable[int]]) -> tf.Tensor:
```
jasonrute, line 376: I'm worried about this code causing problems since it isn't compiled.  Did this already exist in the code base previously?  If not, can we have it produce numpy arrays instead?

jasonrute: Actually, it was already in the code.  I guess it is fine.

jasonrute (approved):
LGTM!  But I think we also really have to test this code with some benchmarks to make sure it still works as intended.

jasonrute (approved):
LGTM again!

PR #118, opened 2023-03-23
Mirek/reindex

jasonrute:
Make the global argument and index into the global context instead of an absolute index.  This will make our model more scalable in the end.

This quickly gets turned back to what it was before, but we will push it further back into the neural network.


jasonrute:
@mirefek, the tests all run (no crashes), but they fail for me.  Can you verify quickly that they pass for you.  (In my upcoming changes I'll reduce the tolerance in hope that our tests will be more robust.)

jasonrute:
Actually, I think this might fix the hmodel. 😄  At least, it makes this line make sense now: https://github.com/IBM/graph2tac/blob/b2779130fff223260b826c03e6c7b28437ce4289/graph2tac/loader/hmodel.py#L34

And it might explain the output of bounds bug I saw in the hmodel last week.  Anyway, let's merge this now, and we can figure out if the hmodel is working in a bit.

PR #120, opened 2023-03-27
Jrute/predict tests

jasonrute:
This makes tests for the predict server and cleans up our other tests.  Highlights:
* There is a single integration test `test_pipeline.py` replacing `test_training_pipeline.py`.
* Broken out testing framework in `pipeline.py` which can easily be used for more targeted tests.
* Added tests for the hmodel.  (But note, the hmodel is actually currently broken, so these tests only check that it runs without crashing and then it is consistent with the current broken model.  One it is fixed we could look into getting a test that tests that it gives the correct answers, besides the usual fingerprint system.  This should be easier now.)
* Added `--replay` argument to `g2t-server` so you can replay coq commands recorded with `--record`.  (This is done in place of creating a TCP or stdin/stdout communication channel with another processes).
* Added tests for the predict server.
  * Predict tests are only in `tests/data/propchain_dep` right now since it covers tactics, local, and global arguments, and it isn't as slow as the mini_stdlib tests.
  * The predict test results are updated with `--overwrite` like the other test results.
  * But `--overwrite` doesn't update the recorded tactician messages.  That only needs to be done when switching dataset versions or other important changes to the communication protocol.  See Jason for now if that is needed.
* Cleaned up many verbose warnings so we can actually see the test output and error messages.
* Switched expected output files (fingerprints) to be yaml.  This will make it easier to see changes in them using diff tools (like github PR file changes)
* Switched all mini-stdlib tests to be one epoch only.  This makes them more likely not to fail on a different machines (since small errors compound over time especially since there is so much data).  (I was having an issue where the tests which pass on Mirek's machine didn't pass on mine.  Hopefully, this cleans this up.)
* All expected result files (which are not newly added files) use the same expected results in `main`, this includes the expected stdlib results for the first epoch.)  They all pass on my machine.
* (Not sure if needed, but) increased error tolerance to 1e-4 relative error (was 1e-5).
* [ ] TODO: Update test documentation.  (I'll do it tomorrow.)
* [ ] TODO (later): I sort of rushed though the creation of the coq record file.  It covers all the Tactician commands, but I could improve it a bit as well.  I should also add a test for `update_new_definitions` (and not just `update_all_definitions`).

Please test it by running `pytest`.  You will need to do `pip install -force-reinstall -e .` first.


convert_json_to_yaml.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
```
LasseBlaauwbroek, line 1: Does this file really need to be here, or was it just used to do the initial conversion? (I don't see it referenced anywhere.)

graph2tac/loader/predict_server.py
```diff
@@ -411,6 +451,7 @@ def prediction_loop(predict_server: PredictServer, context: GlobalContextMessage
                         log_cnts.n_predict += 1
 
                         # important line -- sending prediction
+                        predict_server.response_history.record_response(response)
```
LasseBlaauwbroek, line 454: I thought you were going to write a wrapper for `send`. (Not saying you have to, just wondering why you didn't.)

graph2tac/loader/predict_server.py
```diff
@@ -37,6 +37,44 @@ def apply_temperature(confidences: ArrayLike, temperature: float) -> NDArray[np.
...
 
+class ResponseHistory:
+    """Records response history for testing."""
+    data: dict
+    """JSON compatible dictionary which stores all the history."""
+    def __init__(self, recording_on: bool):
+        """
+        :param recording_on: Record results if True, otherwise do nothing.
+        """
+        self._recording_on = recording_on
```
LasseBlaauwbroek, line 48: It is strange to me to have a recording class that might not be recording. Related to the question below regarding the wrapper for `send`. I would do the logic for recording/not-recording in such a wrapper. (But whatever, I don't care enough about this for an extended discussion.)

graph2tac/loader/predict_server.py
```diff
@@ -433,12 +475,19 @@ def prediction_loop(predict_server: PredictServer, context: GlobalContextMessage
 def start_prediction_loop(predict_server: PredictServer, capnp_socket: socket.socket, record_file: Optional[BinaryIO]):
     prediction_loop(predict_server, capnp_message_generator(capnp_socket, record_file))
 
+def start_prediction_loop_with_replay(predict_server: PredictServer, replay_file: Path):
+    with open(replay_file, "rb") as f:
+        prediction_loop(predict_server, capnp_message_generator_from_file(f))
+
```
LasseBlaauwbroek, line 481: I think that `start_prediction_loop` and `start_prediction_loop_with_replay` might as well be inlined. They are not very useful.

graph2tac/loader/predict_server.py
```diff
@@ -605,15 +654,20 @@ def main():
...
     if config.record is not None:
         record_context = open(config.record, 'wb')
     else:
         record_context = contextlib.nullcontext()
 
     with record_context as record_file:
-        if not config.tcp:
+        if config.replay:
+            assert not config.tcp, "--replay not compatible with --tcp"
+            assert not config.tcp, "--replay not compatible with --record"
```
LasseBlaauwbroek, line 668: The binary invocation for the binaries in graph2tac is already pretty complicated. It would be great if the mutual-exclusivity logic for all this stuff could be encoded into the parser instead of checked after-the-fact. That will make the `--help` information clearer. See [here](https://github.com/coq-tactician/coq-tactician-reinforce/blob/ac300ecf426f94d7a770d0a1dd08dd1e8780493b/pytact/fake_coq_client.py#L32-L42) for an example of that (and also you might look at the `argparse` docs, there are many more options.

tests/data/propchain_dep/params/hmodel/expected_predict_server.yaml
```diff
@@ -0,0 +1,70 @@
+responses:
+- _type: CheckAlignmentResponse
+  contents:
+    unknown_definitions:
+    - <pytact/graph_api.capnp:Definition reader (name = "Tactician.Ltac1.Record.private_constant_placeholder",
```
LasseBlaauwbroek, line 5: If you are going to use yaml, you might as well print this as yaml too. You can get a dictionary version of the data using `d.lowlevel.to_dict()`, which you can then convert to json. Alternatively, you can also just print the node representation (like you do for arguments) using `repr(d.node)`. Or you can print the just the name of the definition using `d.node`.

LasseBlaauwbroek (changes requested):
Great Jason!

The tests pass, but I'm getting this warning:
```
tests/__init__.py:8
  /home/lasse/Documents/Projects/Tactician/develop/graph2tac/tests/__init__.py:8: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: tests
    pytest.register_assert_rewrite('tests')
```

I don't see the `.v` file that you use to generate the trace file. Shouldn't that be included, so that we can update the tests in the future? (I' not saying the updating should be fully automated/streamlined, but just adding the `.v` file to the repo should be easy right?)

LasseBlaauwbroek:
A note about using yaml for test output: Yaml is a notoriously complicated format. (Yes, once parsed it is equivalent to json, but the printed format is ridiculously complicated.) See [the yaml document from hell](https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell). It might be kind of dangerous to use this for tests.

(I would personally probably never use yaml for anything unless it was already part of the system. Instead I would look at toml or something else.)

mirefek:
On the first try, the prediction tests fail for me:

LasseBlaauwbroek:
Did you comply with @jasonrute's instructions?

> Please test it by running pytest. You will need to do pip install -force-reinstall -e . first.

mirefek:
I did

LasseBlaauwbroek:
Including `--force-reinstall?`. Your version of pytactician is out of date.

graph2tac/loader/data_server.py
```diff
@@ -307,7 +307,7 @@ def __init__(self, proportions : list[int], random_seed : int):
         self.random_seed = random_seed
     def lemma(self, d : Definition) -> int:
         # to make it identical to vasily's loader
-        ident_64 = np.uint64(d.node.identity)
```
mirefek, line 310: I think we could actually drop this conversion :-), (while making sure that tests pass) The only purpose of it was to test compatibility with the original loader.

graph2tac/loader/hmodel.py
```diff
@@ -61,7 +60,7 @@ def args_decode(args,
 
 
 class Train:
-    def __init__(self, data_dir: Path, max_subgraph_size, with_context, shuffled, dry):
+    def __init__(self, data_dir: Path, output_dir: Path, max_subgraph_size, with_context, shuffled, dry):
```
mirefek, line 63: Good feature :-), (in my personal tests, I was previously moving `hmodel.sav` around the filesystem :-) )

graph2tac/loader/predict_server.py
mirefek: I am rather unsure if we need two different mechanisms for recording -- `response_history`, and `record_file`. Could we keep only one of them?

graph2tac/loader/predict_server.py
mirefek: Eventually, I was thinking of reorganizing these functions to be methods of `PredictServer` but not necessary to bother with it for now, I think.

mirefek (approved):
Tests pass (even for me), which is I think the important part.

graph2tac/loader/predict_server.py
LasseBlaauwbroek: Mhm, Jason suggested this as well. But if we do this, then we have to generate a new `trace.bin` file whenever something changes in the servers responses. This is possible, I guess. But then there still needs to be logic somewhere that compares an old trace to a new trace.

This is all possible, but it's not clear to me what the API for this should look like. Do you have a suggestion?

jasonrute:
> Great Jason!

Thanks for all the feedback, both of you!

> The tests pass, but I'm getting this warning:
> 
> ```
> tests/__init__.py:8
>   /home/lasse/Documents/Projects/Tactician/develop/graph2tac/tests/__init__.py:8: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: tests
>     pytest.register_assert_rewrite('tests')
> ```
> 
I'll look into it.  I'm not sure if it is easy to fix.  But it isn't breaking anything and I could probably suppress the warning if needed.

> I don't see the `.v` file that you use to generate the trace file. Shouldn't that be included, so that we can update the tests in the future? (I' not saying the updating should be fully automated/streamlined, but just adding the `.v` file to the repo should be easy right?)

Yes, I was holding off until I can make them a bit better, but there is no real reason to.  I can add them to this PR.

convert_json_to_yaml.py
jasonrute: Committed by mistake.  Will remove.

graph2tac/loader/predict_server.py
jasonrute: I was.  I got distracted (and it was a bit more complicated than I thought).  Let me see if I can do it easily.

graph2tac/loader/predict_server.py
LasseBlaauwbroek: In order to facilitate this, and make `--replay` compatible with `--record`, I added a `record` parameter [here](https://github.com/coq-tactician/coq-tactician-reinforce/commit/707fe253cb5a1e72bba95dd7b89878135a9862cc). The relevant signature is now:
```
capnp_message_generator_from_file(message_file: BinaryIO,
                                      check = True,
                                      record: BinaryIO | None = None) -> GlobalContextMessage
```

graph2tac/loader/predict_server.py
jasonrute: @LasseBlaauwbroek I'll see what I can do.

graph2tac/loader/predict_server.py
jasonrute: If easy.  I'll inline them.  If hard, I'll let mirek do it in his refactor.

graph2tac/loader/predict_server.py
jasonrute: Great idea.  I'll do that.  Ideally I would have this as needing exactly one of `--replay`, `--stdin` and `--tcp`, but I that means adding a `--stdin` argument instead of having it be the default behavior if there is no `--replay` or `--tcp`.  @mirefek @LasseBlaauwbroek any objections?

tests/data/propchain_dep/params/hmodel/expected_predict_server.yaml
jasonrute: To be clear, in some sense this is just yaml.  This is just a string.  But yes, I would have like to have it broken up more.  Unfortionately, `d.lowlevel.to_dict()` and other approaches I tried, don't do it automatically.  I get errors about classes without `to_dict`.  So I just stopped at this point and took the representation.  The `repr(d.node())` approach is interesting.  I do have to admit I really like having the name here.  It makes it much easier to see what is going on especially if this test ever breaks at this point.

graph2tac/loader/predict_server.py
LasseBlaauwbroek: You can add the `--stdin` argument, and make it the default can't you? See for example https://stackoverflow.com/questions/38507675/python-argparse-mutually-exclusive-group-with-default-if-no-argument-is-given

graph2tac/loader/predict_server.py
jasonrute: > A note about using yaml for test output: Yaml is a notoriously complicated format. (Yes, once parsed it is a superset of json, but the printed format is ridiculously complicated.) See [the yaml document from hell](https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell). It might be kind of dangerous to use this for tests.
> 
> (I would personally probably never use yaml for anything unless it was already part of the system. Instead I would look at toml or something else.)

That is a good point about yaml.  Maybe in hindsight we shouldn't have invested in yaml for parameters (and went with toml).  But for this particular use case, my main reason is that it makes the test result files readible and diffable.  The later is actually really useful I'm finding.  If we have to change one of these, it is really easy to use git diff to see what we actually changed.  We can't do that with the default printing of JSON, but I guess we could switch to another JSON output format where each piece of data is on its own line.  I just liked YAML since it does it automatically.

And to be clear, the tests are not comparing the files themselves.  They are importing the files into dictionaries and comparing those using pytest's ability to compare dictionaries (and it's ability to compare floats up to a tolerance).

graph2tac/loader/data_server.py
jasonrute: I'm fine with dropping it, but lets do it in it's own PR since it will change all the test results (since it changes the train-test splits).  Since I'm changing the format of the result files, I'd like to at least keep the contents the same so I know I'm not breaking stuff as I go.

graph2tac/loader/predict_server.py
LasseBlaauwbroek: Yes, readable output is a good feature! But in hindsight, I would have used toml, or something else. Whatever, either keep it or switch to toml, up to you. I just thought you should be aware of the pitfalls.

jasonrute: @mirefek The reason for having two recording methods now is as follows.  I originally also just wanted one, but Lasse convinced me this is better and I agree.  Here are what the two recording mechanisms do:
* `--record` with record both incoming and outgoing messages when interacting with coq via stdin or tcp.  It is stored in a binary file.  If I want to make a new recording, I need to have coq running (or fake coq) at least, so I need an external processes running at the same time.  This makes it a bit hard to use in tests where I want to be running our server directly from python and also to be able to update the fingerprints if we change the network.
* `--replay` isn't technically about recording, but about replaying the above record file.  So `--replay` is an alternative to `--tcp` and the default stdin communication channels.  I just currently have it set up to also record outgoing messages only when `--replay` is on to avoid a memory leak were we record all outgoing messages in a long run.  The recording that is happening here is for the tests (although we could use it for other debugging reasons).  I only record the outgoing messages, and I record what I need to know for the tests.  I'm recording to a dictionary which is returned by `main` and can be used in the test file (just like I do for the hmodel and tfgnn train.  It is easy to then do the tests with this dictionary, comparing it to what was previously stored (and comparing the floats up to an error tolerance).  I was hoping I could just do something like `msg.to_dict()` to get the dictionaries for each message, but this is not the case, so it does require a bit more work. Also, I don't have to interact with coq or fakecoq at all here and it is easy to overwrite the tests whenever the network changes from inside `pytest`.

If we had only one system, it would probably look like follows:
* We update the mechanism I use for `--replay` to also allow `--record` to be usable at the same time.  So we read in stored incoming messages from one binary file, and store the new outgoing messages into another.  Lasse could add this functionality if we want.
* Then when we run a test, we don't return the history as a dictionary, but save it to file in the temp directory and the tests read that in.
* The tests would have to read in that record file and convert it to a dictionary for comparison.  So in this sense it would require again writing all that code to convert messages to dictionaries and we wouldn't save much complexity.
* If the tests fail and we use --overwrite, then we replace the old binary record file with this one.  We wouldn't be able to easily see how the test results have changed.

Personally, I like the current approach, and agree with Lasse's original suggestion.

LasseBlaauwbroek: > We update the mechanism I use for --replay to also allow --record to be usable at the same time.

As per a comment of mine earlier in this thread: I just did this.

LasseBlaauwbroek: [One more change](https://github.com/coq-tactician/coq-tactician-reinforce/commit/623952029b0b5e997c08634ba09bb549a0015a0a) to the API, to support @mirefek's preference for a 'unified approach': You can now compare the recorded response with the actual server response through a Callable:
```python
def capnp_message_generator_from_file_lowlevel(
        message_file: BinaryIO,
        check : Callable[[Any, Any, Any], None] | None = None) -> (
        Generator[pytact.graph_api_capnp_cython.PredictionProtocol_Request_Reader,
                  capnp.lib.capnp._DynamicStructBuilder, None]):
```
The callable of `check` will be called like
```python
check(request_message, server_response_message, recorded_response_message)
```
I believe that this is enough to implement the 'unified approach'. Wether or not to actually do that, I'll leave for @mirefek and @jasonrute to discuss.

mirefek: Thanks! I would rather not spend time on it with this commit, and get back to it in a later refactor of `predict_server` (say when I would get to work on nested contexts).

tests/data/propchain_dep/params/hmodel/expected_predict_server.yaml
jasonrute: @LasseBlaauwbroek I don't suppose there is a similar innovation to get the name of the tactic or the name of the local argument, is there?  That would make the output a lot more readable.

LasseBlaauwbroek: Regarding local arguments: I just [pushed](https://github.com/coq-tactician/coq-tactician-reinforce/commit/413caa73aa607897f4ac9dbcabf911c3fa1cc161) a change with a new option `Set Tactician Neural Metadata`, which will force Coq to include all of the metadata that is also present in the dataset when transmitting data to the server. This can be pretty expensive, so you should only use it in test/debug situations. The name of a local argument `a : Node` can be found by finding `i` such that `proof_state.context[i] == a` and then finding its name using `proof_state.context_names[i]` and optionally it's type using `proof_state.context_text[i]`. Here `proof_state` refers to to object being cast on line 435:
```
if isinstance(msg, ProofState)
```

Regarding tactic names: Currently, the capnproto format doesn't allow for that. But we could certainly add this as an option in the future (please open an issue if you want that).

LasseBlaauwbroek: (And yes, I know that this method of looking up names of local context elements is cumbersome. But this cannot be helped due to the graph sharing algorithm. Structurally equal hypotheses that are part of different proof states will be shared using the algorithm. But they can still have different names. So the node corresponding to a hypothesis doesn't have a unique name. Therefore, the name has to be stored in the proof_state object, where it is unique.

jasonrute:
@LasseBlaauwbroek If I put the `.v` files used to generate the recordings under [`tests/data/propchain_dep`](https://github.com/IBM/graph2tac/tree/main/tests/data/propchain_dep) or in a subdirectory, will they get captured when running this [MakeFile](https://github.com/IBM/graph2tac/blob/main/tests/data/propchain_dep/Makefile)?

LasseBlaauwbroek:
Only if you put them under `tests/data/propchain_dep/theories/`. But this is probably not what you want right? Because generating the dataset is quite different from generating a trace (requiring a different branch of `coq-tactician-reinforce` to be installed). So at least for now, they should probably have a different Makefile (or a different target within the same makefile).

LasseBlaauwbroek:
For example, you can add something like this to the makefile:
```
trace:
	tactician exec coqc my_coq_file.v
```
which would then be called like `make trace`.

jasonrute:
@LasseBlaauwbroek I'm done making revisions.  I should have addressed your concerns mostly, with the following exceptions which I didn't get around to but could do later in another PR:
* [ ] Didn't get to putting the recording class code under some `.send` wrapper
* [ ] I didn't look into toml as an alternative to yaml
I also feel this still has the following issues:
* [ ] Didn't do anything with local arguments (as you mentioned was now possible)
* [ ] I don't particularly like the .v files I used to generate traces
* [ ] Didn't update the testing documentation.  That might be a bit out of date now.
* [ ] I think we still have some disagreement about if we want two ways to record what is going on with coq.

PR #121, opened 2023-03-29
Mirek/add split

mirefek:
Only adding the code / splits so that we don't have to search for it in Slack's history. The code is just how I was running it, so it would need more tweaks if we want to make it an actual (command line) tool. Would you like me to do it?


LasseBlaauwbroek (approved):
Thanks. It's fine for now I think. We can always improve more later. But it's not like we will be re-doing this every week, so no hurry I think.

PR #122, opened 2023-03-30
more flexible tactic uniqueness filter

mirefek:
`exclude_unique_tactic: false` ->`required_tactic_occurrence: 1`
`exclude_unique_tactic: true` ->`required_tactic_occurrence: 2`

tests pass

### DIFF
--- a/graph2tac/loader/data_server.py
+++ b/graph2tac/loader/data_server.py
@@ -390,7 +390,7 @@ class DataServer(AbstractDataServer):
             file_data = self._data[name]
             self._load_file(file_data)
 
-        if self.dataset_config.exclude_unique_tactics:
+        if self.dataset_config.required_tactic_occurence > 1:
             self._exclude_unique_tactics()
 
         # precalculate def_file_ctx in a forward order to prevent stack overflow,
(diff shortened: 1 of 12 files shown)


graph2tac/loader/data_server.py
```diff
@@ -390,7 +390,7 @@ def __init__(self, data_dir : Path, dataset_config : DatasetConfig):
             file_data = self._data[name]
             self._load_file(file_data)
 
-        if self.dataset_config.exclude_unique_tactics:
+        if self.dataset_config.required_tactic_occurrence > 1:
```
LasseBlaauwbroek, line 393: Maybe just move this check into `_exclude_unique_tactics`?

LasseBlaauwbroek (approved):
Looks good. Thanks. I'll let @jasonrute merge when he approves.

jasonrute (approved):
I'm approving since it isn't a big deal, and I'm fine with you merging as is, but I confused about having 1 be the lowest number, or am I mistaken and there is a difference between 0 and 1?  Is the idea that 0 means keep all tactics in the dataset even if never used, and 1 means take only tactics used at least once?  (And it just happens that our dataset has no tactics used zero times, because why would it?)

mirefek:
There is not difference between 0 and 1.

mirefek:
It was a bit easier to phrase the argument positively: "keep only tactics occuring at least k-times", and there is indeed no way to detect a tactic occuring 0 times.

jasonrute:
Ok.  I'll merge it.

PR #123, opened 2023-03-30
Avoid dependence on valid_indices

jasonrute:
This simple hack makes it so that the inference model returns logits according to the dynamic global context.  It makes the following assumptions (which at least are currently true):
* the dynamic global context are indices into the valid_indices
* the batch size to the inference model is 1 (or at least the dynamic global indices are the same size for all elements of the batch)

To make sure this works, I made another branch [`jrute/avoid-dependence-on-valid-indices-with-tests](https://github.com/IBM/graph2tac/tree/jrute/avoid-dependence-on-valid-indices-with-tests) where I did two things:
* Merged in the predict server tests and made sure they passed
* Made an [experimental commit](https://github.com/IBM/graph2tac/commit/10ed4d18f9f76ad5fd424d1a10a20df5a0ba78dc) where I messed with both the valid_indices and the dynamic global context and made sure it still passed the tests as intended.

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -638,6 +638,14 @@ class LocalArgumentPrediction(TacticPrediction):
         num_arguments = tf.shape(logits)[1]
         num_logits = tf.shape(logits)[2]
         return tf.reshape(logits, shape=(tactic_expand_bound, -1, num_arguments, num_logits))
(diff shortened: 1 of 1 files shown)


mirefek:
Thanks, I will try it tomorrow.

mirefek (approved):
I am merging to reduce the number of PR's, also so that it will get soon merged with predict-tests

PR #124, opened 2023-03-30
Proposal for better argument parse

LasseBlaauwbroek:
PR-on-top-of-a-pr. Why not do the argument parsing like this? I also think we should get rid of `--hostname` and `--port` (and a bunch of other things), but that's for another time.


LasseBlaauwbroek:
To move things along, I'll just merge this. I don't think there is any reasonable objection to make here.

PR #126, opened 2023-04-03
Jrute/fix accuracy calc

jasonrute:
This PR fixes a subtle issue in the strict and sequence accuracy calculations.  The logits are currently stored in an array of shape `[batch_size, max(num_args), context]` where `max(num_args)` is the maximum number of args in the batch, and `context` is all the global definitions.  It made the following assumption about argument positions which were not used for that give tactic: the argmax of the logits would be the 0th position.  This happened to be true but was easy to break by changing all manner of otherwise non-essential details in the calculation.  This PR fixes that issue by removing those positions from the accuracy calculation altogether.  (Note.  In future PRs we will also remove a dependence on logits not in the global context, but this PR doesn't cover that yet.  This PR was a simple PR to address broken tests in another branch.)

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -119,7 +119,7 @@ class ArgumentSparseCategoricalCrossentropy(tf.keras.losses.Loss):
         super().__init__(**kwargs)
         self.sum_loss_over_tactic = sum_loss_over_tactic
 
-    def arguments_filter(y_true: tf.RaggedTensor, y_pred: tf.Tensor) -> Tuple[tf.RaggedTensor, tf.RaggedTensor]:
+    def arguments_filter(self, y_true: tf.RaggedTensor, y_pred: tf.Tensor) -> Tuple[tf.RaggedTensor, tf.RaggedTensor]:
         """
         Extracts the local arguments which are not None from the ground truth and predictions.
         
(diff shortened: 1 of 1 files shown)


jasonrute:
It doesn't yet deal with different context sizes.  Instead, it uses the old behavior where for local arguments, the context size is rounded to the max local context size (and all extra values get logits of -inf.  Similarly, for globals, the context size is constant (and instead a masking mechanism is used to account for the global context size, where all context elements not in the dynamic global context get a logit value of -inf).

Instead the point of this PR is to deal with the global argument positions which are not used for a tactic.  (If you are curious, they currently work as follows.  At some point that argument dimension is converted from ragged to non-ragged in the query vector and is filled in with zeros.  Hence in the linear operation, the query-key dot product is zero for all logits in that local argument position, except those which are masked as -inf.  Previously, we could always count on the first global argument logit being zero and not -inf, so the argmax for global args for such unused argument positions would be 0.  But that is a dangerous assumption.  So I just removed some argument positions from the calculation.)

I am ultimately redoing this calculation soon where the context dimension will also be ragged, so I guess it isn't really necessary to document this too much, but if it is still confusing after that point, maybe I should figure out how to comment the code better.

mirefek:
I see, I think it is fine, we can merge

jasonrute:
Also, I just noticed a bug.  If the model picks a local argument with say index 5 but the correct answer is the global argument with index 5 then the current code I just wrote will mark that as accurate since it only looks at the index.  I need a bit more to make sure we made the right local-global choice.  Give me one moment to fix this PR before merging it.

mirefek:
ok

jasonrute:
Fixed the bug.

PR #127, opened 2023-04-04
Mirek/reindex2

mirefek:
Indexing global context from 0 instead of 26 -- a step towards getting rid of the static global context


graph2tac/loader/predict_server.py
```diff
@@ -285,10 +284,14 @@ def _enter_coq_context(self, definitions: OnlineDefinitionsReader, tactics):
 
         self.data_server.align_definitions(definitions.definitions())
 
-        logger.info(f"initializing network with {len(self.data_server.global_defs)} defs in global context")
+        if len(self.data_server.global_defs) > 0:
+            static_global_context = np.arange(max(self.data_server.global_defs)+1, dtype=np.uint32)
+        else:
+            static_global_context = np.arange(1, dtype=np.uint32)
+        logger.info(f"initializing network with {static_global_context} defs in global context")
```
jasonrute, line 291: Should this be `{len(static_global_context)}` like it used to be.  I think for an info-level logging statement, listing all the indices in the global context would be too much.

tests/data/propchain_dep/params/hmodel/expected_predict_server.yaml
jasonrute, line 1: It's really interesting (as you pointed out to me privately) that this PR seems to fix our issues with the hmodel, at least in this case it is giving the predictions it should be giving.

jasonrute (approved):
LGTM!  One small issue with a logging statement, but otherwise looks good.  As we talked about, let's merge after #126 is merged into main and we fix the test outcomes to go back to what is in main (except the hmodel outcomes).

PR #129, opened 2023-04-04
Use ragged tensors for training model

jasonrute:
This PR is one initial step into speeding up and improving memory efficiency in the model.  The major changes are as follows:
* The training model now outputs a tensor of shape `[batch_size, None(num_args), None(context)]`.  In other words the arguments and context dimension are now ragged where the arguments dimension only ranges over argument positions used for that tactic, and the context dimension only ranged over elements in the local context or dynamic global context.
* The ground truth global arguments given to the model are indexed into the dynamic global context now.  This matches the way they are outputted by the training model.
* The losses and metrics are changed accordingly.  There are still some inefficiencies (especially in the accuracy calculations) where I convert a ragged tensor of logits back into a non-ragged tensor.  That can be fixed a bit later, but I want to start working on the main bottleneck, which is the linear operation.

This PR is a first step to speeding up the model.  Now that we have a foothold with the ragged tensor, future PRs can address the following:
- [ ] Making the linear operation (matrix multiplication) which creates the logits more efficient (but only performing the calculation over the dynamic global context and the valid argument positions)
- [ ] Consolidating the code for the local and global arguments since they are now using the same idea
- [ ] Remove any dependencies on `_valid_indices` which is no longer needed.
- [ ] Clean up the accuracy calculations to not convert back to non-ragged tensors first.
- [ ] Clean up the prediction calculation to not convert back to non-ragged tensors first.

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -4,6 +4,7 @@ import yaml
 import tensorflow as tf
 import tensorflow_gnn as tfgnn
 from pathlib import Path
+import numpy as np
 
 from graph2tac.loader.data_classes import GraphConstants
 from graph2tac.tfgnn.graph_schema import proofstate_graph_spec, batch_graph_spec, strip_graph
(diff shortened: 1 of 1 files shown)


jasonrute:
The tests all pass (and were very helpful in finding some subtle bugs).

jasonrute:
Yes.  The output of the prediction network hasn't been changed in this PR.  I'm slowing modifying everything systematically.

PR #130, opened 2023-04-17
Remove valid_indices

jasonrute:
This simple PR removes the need for valid_indices, which was already superfluous since it was always equal to `range(n)` where `n` was the length of the embedding tensor, and hence `tf.gather(emb_matrix, self._valid_indices)` was a no-op.

The tests all pass.


mirefek (approved):
Looks good, thanks. Eventually, we could get to removing `graph_constants.global_context` but I think this can be already merged.

PR #131, opened 2023-04-19
Speed up model by switching to ragged tensors

jasonrute:
This is the second PR in a series of PRs to speed up the model and reduce the memory requirements.  It refactors major parts of the argument prediction code, and is enough to start training on a large dataset.

The tests all pass.  (A unit test was added for the new `QueryKeyMul` layer which multiplies the queries and keys into logits.  This was an expensive computation and it required a lot of optimization.  This ensures the optimizations do the write calculations, and lets us even further optimize it if needed.)

Work to still be done in future PRs.  Probably none of these are necessary to train and benchmark the model, so it might be enough to start large runs now.  But here are some possible things we could do before that.  Note, if we do this later, we might have to retrain a model to be compatible with the new code.  We should add some tests that check for compatibility with already training models to know if we are making breaking changes.
- [ ] Make the prediction network return a ragged tensor (and refactor the prediction code to use that tensor instead).  This would likely speed up some parts of the predict server.
- [ ] Optimize the prediction network for speed/memory.
- [ ] Delete useless code (can be done at any time and isn't a blocker)
- [ ] Streamline some of the code.  (Currently, this new code is a bit convoluted at parts, with layers calling layers, etc.  It would be good to clean it up, but not needed.)

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -665,17 +665,35 @@ class GlobalEmbeddings(tf.keras.layers.Layer):
 
 
 class QueryKeyMul(tf.keras.layers.Layer):
+    """Ragged Query-Key multiplication
+
+    Calculate inner product of ragged key and query tensors.
+    It is one of the most computationally intensive in the network and hence has been heavily
+    optimized.
+
+    :param method: One of multiple methods to computing the query and key.
+    :type tf: str
+    :param name: Layer name
+    :type tf: str
+    """
+
     def __init__(
         self,
-        method="broadcast_ragged2",
-        name="query_key_mul",
+        method : str = "broadcast_ragged",
+        name : str = "query_key_mul",
         **kwargs
     ):
         super().__init__(name=name, **kwargs)
         self.method = method
 
     @staticmethod
-    def _mul_map_fn(queries, keys):
+    def _mul_map_fn(
+        queries : tf.RaggedTensor,  # shape: # [batch, None(args), hdim]
+        keys : tf.RaggedTensor,  # shape: # [batch, None(context), hdim]
+    ) -> tf.RaggedTensor:  # shape: # [batch, None(args), None(context)]
+        """
+        Compute the inner product via tf.map_fun.  Won't run on a GPU.
+        """
         @tf.function
         def linear_op(qk):
             x = tf.einsum("ij, kj -> ik", qk[0], qk[1])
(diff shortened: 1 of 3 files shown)


graph2tac/tfgnn/tasks.py
```diff
@@ -615,6 +646,228 @@ def metrics() -> Dict[str, List[tf.keras.metrics.Metric]]:
...
+        self, 
+        global_context: tf.RaggedTensor,  # [batch, None(context)]
+        inference: bool = False,
+        training=False
+    ) -> tf.Tensor: # [batch, None(context), hdim]
+        # [valid_context, hdim]
+        all_embeddings = self.global_embeddings_layer.get_keys_embeddings()
+
+        # note: inference model always uses dynamic global context
+        if self.dynamic_global_context or inference:
```
mirefek, line 671: Is there still any meaning in `self.dynamic_global_context = False`? Perhaps, we could discard this option.

mirefek (approved):
I suppose that you checked the standard tests, and also the special QueryKeyMul test with all four computation options, right? Then I am fine with merging.

graph2tac/tfgnn/tasks.py
jasonrute: We could still support it.  We would need to convert the full global context to the dynamic one right before the output, but for training purposes it would do the multiplication on all the definitions.  But also, I don't see the point, so maybe we should just kill this option to simplify the code.  (It is broken currently, anyway.)

graph2tac/tfgnn/tasks.py
```diff
@@ -64,6 +64,18 @@ def ragged_logits(y_true: tf.RaggedTensor, y_pred: tf.Tensor) -> Tuple[tf.Ragged
...
+
+    # expand context (replacing unused context element logits with -inf)
+    # shape: [batch, None(args), max(context)]
+    logits = logits.with_values(logits.values.to_tensor(default_value=-np.inf, shape=[None, context_size]))
+
+    # expand args (replacing unused arg position logits with 0.0)
+    # shape: [batch, max(args), max(context)]
+    logits = logits.to_tensor(default_value=0.0)
+
+    return logits
```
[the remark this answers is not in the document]
jasonrute: This function is currently only used in the inference model to convert the output to a dense tensor.  (It is not used in the multiplication.) The plan is for it to be temporary yes.

jasonrute:
> I suppose that you checked the standard tests, and also the special QueryKeyMul test with all four computation options, right? Then I am fine with merging.

The new unit test for QueryKeyMul checks that all the multiplication methods come out the same.  I've also manually checked (at one point) that the integration tests work with all the methods, but this isn't a regular part of integration testing.  I figure the unit test makes it so that we don't have to run the whole pipeline on the other methods.

Note, one of those multiplication methods, namely "map_fn", doesn't work on a GPU.  So even though it works on a CPU and gets the right answer, we couldn't use it for GPU training (but maybe for CPU inference if it is faster).

PR #132, opened 2023-04-22
Add dataset with duplicated definitions

LasseBlaauwbroek:
Per request of @mirefek. Trace file is still missing due to this failure. Any suggestions of what is causing that?

```
ValueError: Exception encountered when calling layer "tf.math.top_k" (type TFOpLambda).

input must have last dimension >= k = 3 but is 2 for '{{node tf.math.top_k/TopKV2}} = TopKV2[T=DT_FLOAT, sorted=true](Placeholder, tf.math.top_k/TopKV2/k)' with input shapes: [?,2], [] and with computed input tensors: input[1] = <3>.

Call arguments received by layer "tf.math.top_k" (type TFOpLambda):
  • input=tf.Tensor(shape=(None, 2), dtype=float32)
  • k=3
  • sorted=True
  • name=None
```


LasseBlaauwbroek:
Okay, I found the issue. This exception happens when the amount of available tactics is lower than the tactic expand bound. This edge-case should probably be fixed!

Trace added.

LasseBlaauwbroek:
Btw, the amount of boilerplate that is needed for a new test is a bit much. Can this be improved?

mirefek (approved):
The test works for me, thanks.

LasseBlaauwbroek:
Updated: I made some mistake that caused no definitions to be aligned. That's probably not what we want. Fixed now.

jasonrute:
Let me just run this on my machine and if it passes fine, I'll merge it.

jasonrute:
It passes for me.  It takes 2 minutes and 45 seconds which is quicker than the other tfgnn server tests, but still longer than it probably should be.  But we can work on that later.  Merging...

PR #133, opened 2023-04-24
Mirek/predict nesting

mirefek:
It is not finished yet but quite close to it. What remains:
- [x] Merge Lasse's test `duplicate_identities`
- [ ] Merge fixed `hmodel` test
- [ ] Test on these tests, fix hmodel, possibly other bugfixes
- [x] Delete ad-hoc debug prints
- [x] Fix `LoggingCounters`
- [x] Set `paranoic = False` by default
- [x] Try the PR with updated version of `coq-tactician-reinforce`

so this is now waiting for the tests being merged into `main`. (or I could first merge Lasse's branch directly if it will take a while)

### DIFF
--- a/graph2tac/predict.py
+++ b/graph2tac/predict.py
@@ -90,7 +90,7 @@ class Predict:
         """
         [ Public API ] Prepares sufficient size for (new) model's definition
 
-        @param global_context: a replacement for the original global_context in the GraphConstants seen during training
+        @param new_node_label_num: required size of the array of nodes stored in the model
         """
         raise NotImplementedError('allocate_definitions should be implemented by sub-classes')
 


split/testing.txt
```diff
@@ -1,18 +1,18 @@
...
+coq-tlc.20200328
+coq-iris-heap-lang.3.4.0
+coq-printf.2.0.0
+coq-smtcoq.2.0+8.11
+coq-topology.10.0.1
+coq-haskell.1.0.0
+coq-bbv.1.3
+coq-poltac.0.8.11
+coq-mathcomp-odd-order.1.14.0
+coq-hott.8.11
```
LasseBlaauwbroek, line 18: What is happening here?

graph2tac/loader/predict_server.py
```diff
@@ -426,63 +564,46 @@ def prediction_loop(predict_server: PredictServer, context: GlobalContextMessage
...
-            with predict_server.coq_context(context):
-
-                log_cnts.msg_idx = -1
-                log_cnts.t_predict = 0.0
-                log_cnts.n_predict = 0
-                log_cnts.max_t_predict = 0.0
-                log_cnts.argmax_t_predict = -1
-                log_cnts.t_coq = 0.0
-                log_cnts.n_coq = 0
-                t0_coq = time.time()
```
LasseBlaauwbroek, line 445: Looks like the logging counters are broken?

LasseBlaauwbroek (commented):
Some comments, but I'm not seeing any major things wrong. (Not tested it yet, just looked at the code; also I don't fully understand all of the code.)

I'm seeing quite a bit of outcommented code. Do you want to keep that? Also, is the `sanity_check` function expensive? If so, should it be behind a flag?

graph2tac/loader/predict_server.py
mirefek: Oh, you are right, thanks for the remainder, I was planning to look at them later. I will add this to the TODO list at the top.

split/testing.txt
mirefek: Converting CRLF -> LF (this is a little unimportant fix aside the main purpose of the branch indeed)

mirefek:
> I'm seeing quite a bit of outcommented code. Do you want to keep that?

I will remove the outcommented debug prints.
> Also, is the `sanity_check` function expensive? If so, should it be behind a flag?

`sanity_check` is a functionality that was there also before, and I hope it should not be too expensive, so I will keep it. On the other hand, I plan to set `paranoic = False` as default for the `DynamicDataServer`.

LasseBlaauwbroek:
(Sorry, I didn't read your comment about deleting debug code.)

mirefek:
The code is already in virtually finished state. The only missing part is a functional hmodel test -- I already updated the hmodel code, and tested it (thanks, Jason) on the `duplicate_identities` test. Unfortunatelly, this test doesn't contain any global argument prediction so I cannot yet guarantee that it is working.
On the other hand, since the only possible planned future changes are bugfixes, the code is ready to be thoroughly reviewed, @jasonrute , @LasseBlaauwbroek

mirefek:
Works about the same with the updated version of `coq-tactician-reinforce`, I also tested with `Set Tactician Neural Autocache.` and looks to me as expected.

LasseBlaauwbroek:
Great, no more problems with the bugs that existed before?

jasonrute:
I'm still working through all the code.  It is fairly complicated.  The big thing I'm confused by is the following line:
https://github.com/IBM/graph2tac/blob/bbe77e619a1f1043e95abf36c305d3b4e4b6abdf/graph2tac/tfgnn/predict.py#L227
If `reserve` is 0.5 (which is the default), doesn't this cause the array to shrink?  And that isn't what we want.  I've very confused by how this works and pasts our tests.

mirefek:
@LasseBlaauwbroek : I haven't encountered any issues.
@jasonrute Yes :-). First I had there only multiplication but then it seemed to me that adding this is more fitting the parameter name `reserve`.

jasonrute (approved):
I've looked over this.  I have a few comments, nothing big.  Honestly, I wasn't able to make it through all the `predict_server.py` code.  The logic is fairly complicated.  But I'll assume it works, and I'll also do some more direct testing as I work to speed up the model.  Also, I like that more of the logging stuff got moved into the logger class.  After addressing my small points, I'm happy to have you merge it.

graph2tac/loader/predict_server.py
```diff
@@ -177,8 +232,9 @@ def summary_log_message(self) -> str:
         )
 
 class DynamicDataServer(AbstractDataServer):
-    def __init__(self, graph_constants: GraphConstants):
+    def __init__(self, graph_constants: GraphConstants, paranoic = False):
```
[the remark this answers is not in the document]
mirefek: Yes, it is mainly for debugging to make sure it doesn't break inside. We could turn it on for tests but I don't know how tests work. I can make it a command line parameter.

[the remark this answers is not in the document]
jasonrute: Yes, make it a command line parameter.  Then let's just add it to all the predict server tests.  Do that by adding the command line parameter to the relevant `predict_server.yml` files, e.g. https://github.com/IBM/graph2tac/blob/main/tests/data/propchain_dep/params/predict_server_update_all/predict_server.yml
(If say the parameter is `--foo` (with no arguments), then add `foo: null` to the file.)

jasonrute:
Great.  Looks good to me!  (And thanks for fixing the testing docs.).

PR #134, opened 2023-04-24
Upgrades to the tests

jasonrute:
This PR upgrades the tests in various ways:
* Added tests which checks consistency with a previous version of the model.  This version is based on the network code currently in main and unless we plan to retrain the model, we should try to keep the test passing.
* Cache training of models in the training tests (which speeds up the predict server tests since they don't necessarily have to retrain a model first).
* Predict server tests can share the results from the same training test.
* Split the different steps of the pipeline into their own named tests so that once can easily say run all the server tests or all the hmodel tests.
* Added a predict_server test which doesn't update names.
* Updated the test documentation.  (Please look it over and see if it makes sense.  I won't go into too many details here in hope that you can get those details from the documentation.)


jasonrute:
Note this PR doesn't do some of the other stuff we are planning to do with the tests, like align definitions in the traces, speed up the predict server tests by making them shorter, and a number of other small improvements.  I'll work on those next.

LasseBlaauwbroek (approved):
Looks like an improvement to me. Tests (23) pass on my machine. Some warning though (which were always there I think):

```
================================================== warnings summary ==================================================
venv/lib/python3.10/site-packages/tensorflow/python/framework/dtypes.py:205
  /home/lasse/Documents/Projects/Tactician/develop/graph2tac/venv/lib/python3.10/site-packages/tensorflow/python/framework/dtypes.py:205: DeprecationWarning: `np.bool8` is a deprecated alias for `np.bool_`.  (Deprecated NumPy 1.24)
    np.bool8: (False, True),

venv/lib/python3.10/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:326
  /home/lasse/Documents/Projects/Tactician/develop/graph2tac/venv/lib/python3.10/site-packages/tensorboard/compat/tensorflow_stub/dtypes.py:326: DeprecationWarning: `np.bool8` is a deprecated alias for `np.bool_`.  (Deprecated NumPy 1.24)
    np.bool8: (False, True),

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

mirefek (commented):
I also have a bit trouble understanding how the `--retrain` should work. I will ask you via Slack.

PR #138, opened 2023-05-02
Quick speedup to inference

jasonrute:
This is a very simple change which speeds up the inference part of the predict server.  Indeed 80% of the computation spent in the neural network for inference (on CPU) was in this one line for some weird reason.  The new line is equivalent, and I don't understand why it is so much faster on the CPU. 🤷 

For my test cases this allows us to do say 600 messages per minute instead of 500.

(Note, the 80% above refers to the neural network part running the inference model.  There is a significant amount of time, much more than the neural network, spent calculating the highest scoring results based on the networks outputs.  My next step is to tackle that part of the code.)

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -670,8 +670,10 @@ class GlobalEmbeddings(tf.keras.layers.Layer):
         # note: inference model always uses dynamic global context
         if self.dynamic_global_context or inference:
             # select the global context for each batch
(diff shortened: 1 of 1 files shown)


jasonrute:
One test fails, namely `tests/integration/test_pipeline.py::test_predict_server[propchain_dep-hmodel-predict_server_hmodel]`, but this is expected, right @mirefek?

mirefek:
yes, it is failing after my update

mirefek (approved):
Looks good, I am surprised that RaggedTensors don't do this automatically right.

PR #139, opened 2023-05-03
Mirek/predict control compile

mirefek:
Not waiting with compilation of the predict server to a prediction, rather compiling at the `__init__` of `Predict`. I wanted to also fully compile `compute_and_replace_definition_embs` but there is a problem with creating a working dummy empty input for this task, and it is less crucial to have this task precompiled. I also returned the `build_network_time` to the `LoggingCounters`.


mirefek:
all tests pass except the known hmodel prediction

graph2tac/tfgnn/tasks.py
```diff
@@ -1271,8 +1271,7 @@ def create_inference_model(self, tactic_expand_bound: int, graph_constants: Grap
 
         # [batch_size, ]
         no_local_context_proofstates = scalar_proofstate_graph.context['local_context_ids'].row_lengths() == 0
-        no_global_context_proofstates = tf.fill(dims=(proofstate_graph.total_num_components,),
-                                                value=len(graph_constants.global_context) == 0)
+        no_global_context_proofstates = scalar_proofstate_graph.context['global_context_ids'].row_lengths() == 0
```
jasonrute, line 1274: Why do we have this?  When would it come up that there is literally nothing in the global context?  I think even tests have stuff in the global context, right?  But since it is in the code, we might as well just leave it for now.  Thanks for the fix.

graph2tac/loader/predict_server.py
```diff
@@ -138,6 +140,13 @@ class LoggingCounters:
     May incorrectly be 0.0 if there isn't enough time between 'initialize' calls.
     """
 
+    @contextmanager
+    def measure_build_network_time(self):
+        t0 = time.time()
+        yield
+        t1 = time.time()
+        self.build_network_time += t1 - t0
```
jasonrute, line 148: Is this ever reset, or is this the cumulative time since the start of the server?

jasonrute (approved):
LGTM!  Just a few small questions and comments.

graph2tac/tfgnn/predict.py
```diff
@@ -290,6 +271,66 @@ def _make_proofstate_graph_tensor(self, state : LoaderProofstate):
...
+                global_context = np.zeros([0], dtype = int),
+            ),
+            metadata = ProofstateMetadata(
+                name = "",
+                step = 0,
+                is_faithful = False,
+            ),
+        )
+        # TODO?: make a working dummy input for _compute_and_replace_definition_embs
+        # currently, empty list of updated definitions crashes in the name_layer
```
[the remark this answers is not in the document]
mirefek: I agree, it would be a bit nicer to also precompile this but it is not a big deal.

graph2tac/tfgnn/tasks.py
mirefek: I don't know either, I agree it is probably a relict, I was just not fully sure

graph2tac/loader/predict_server.py
mirefek: Cumulative, I never reset this counter. The number of network recompilations should be little.

PR #140, opened 2023-05-04
measure time of prediction_requests.send()

mirefek:
Also measuring the `prediction_requests.send()`. On the other hand, it didn't seem to me that there would be a significant time difference from the other coroutine in my simple tests. Maybe, @jasonrute could check with his experiments.

### DIFF
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -110,9 +110,11 @@ class LoggingCounters:
     """Max time (in seconds) spent on single predict since last 'initialize'"""
     argmax_t_predict: int = -1
     """Index of last maximum time single predict since last 'initialize'"""
-    t_coq: float = 0.0
+    t_coq_send: float = 0.0
     """Total time (in seconds) spent waiting on Coq since last 'initialize'"""
-    t0_coq: float = -1.0
+    t_coq_receive: float = 0.0
+    """Total time (in seconds) spent waiting on Coq since last 'initialize'"""
+    t0_coq_receive: float = -1.0
     """last (absolute) time we called coq"""
     n_coq: int = 0
     """Number of 'prediction' requests made by Coq"""
@@ -155,18 +157,23 @@ class LoggingCounters:
         self.n_def_clusters_updated += n_def_clusters_updated
         self.update_def_time += t1 - t0
 
-    def measure_t_coq_start(self):
-        if self.t0_coq >= 0:
+    @contextmanager
+    def measure_t_coq_send(self):
+        t0 = time.time()
+        yield
+        t1 = time.time()
+        self.t_coq_send += t1-t0
+        if self.t0_coq_receive >= 0:
             logger.warning("(coq_start) Nesting theorems in LoggingCounters, likely causing to misleading summaries")
-        self.t0_coq = time.time()
+        self.t0_coq_receive = t1
     def measure_t_coq_finish(self):
-        t1_coq = time.time()
-        if self.t0_coq < 0:
+        t1_coq_receive = time.time()
+        if self.t0_coq_receive < 0:
             logger.warning("(coq_finish) Nesting theorems in LoggingCounters, likely causing misleading summaries")
             return
         self.n_coq += 1
-        self.t_coq += t1_coq - self.t0_coq
-        self.t0_coq = -1.0
+        self.t_coq_receive += t1_coq_receive - self.t0_coq_receive
+        self.t0_coq_receive = -1.0
 
     @contextmanager
     def measure_t_predict(self):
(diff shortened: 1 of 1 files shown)


graph2tac/loader/predict_server.py
```diff
@@ -609,10 +618,9 @@ def prediction_loop(self, context: GlobalContextMessage):
 
                     self.response_history.record_response(response)
 
-                    # sending response back to coq
-                    context.prediction_requests.send(response)
-
-                    self.log_cnts.measure_t_coq_start()
+                    with self.log_cnts.measure_t_coq_send():
+                        # sending response back to coq
+                        context.prediction_requests.send(response)
```
jasonrute, line 623: While I like the context manager approach, I don't think it is good to basically put the old `measure_t_coq_start()` inside the context manager for something else.  I think leaving `measure_t_coq_start()` as it's own thing is more clear.

mirefek: ok, I will make it separate

graph2tac/loader/predict_server.py
```diff
@@ -155,18 +157,23 @@ def measure_update_def_time(self, n_def_clusters_updated):
...
-        self.t0_coq = time.time()
+        self.t0_coq_receive = t1
     def measure_t_coq_finish(self):
-        t1_coq = time.time()
-        if self.t0_coq < 0:
+        t1_coq_receive = time.time()
+        if self.t0_coq_receive < 0:
             logger.warning("(coq_finish) Nesting theorems in LoggingCounters, likely causing misleading summaries")
             return
         self.n_coq += 1
```
jasonrute, line 174: Isn't there an off-by-one error with using the same counter for receive and coq_send.  We will send one more message than we receive, right (since we aren't counting the time to receive the original message)?

jasonrute:
@LasseBlaauwbroek We are still recording the iterator time I think, but it was hidden.  See my comment above.

graph2tac/loader/predict_server.py
mirefek: One of the reasons I put it inside was actually that I wanted just a single counter `n_coq`. There is also an alternative to get rid of that counter entirely and use just the step index.

graph2tac/loader/predict_server.py
mirefek: I actually also count receiving the "iteration termination" message

LasseBlaauwbroek:
Okay, yes, I still misunderstood the code. LGTM.

Another drive-by-comment: Aren´t `LoggingCounter.msg_idx` and `LoggingCounters.n_predict` the same thing? Can one be removed?

mirefek:
Yes, I think it will probably be better to keep just one counter

LasseBlaauwbroek:
In this particular case, these variables seem to have exactly the same semantics. Their docs is also (virtually) the same. I don't see any advantage to keeping them separate.

mirefek:
They are not completely the same, `msg_idx` is initialized to `-1`, the others to 0.

mirefek:
I tried to address the comments, I hope I haven't introduced a +-1 error

graph2tac/loader/predict_server.py
```diff
@@ -230,15 +224,12 @@ def summary_log_message(self) -> str:
             "Initialize|Network build time (s)" : f"{self.build_network_time:.6f}",
             "Initialize|Def clusters to update" : f"{self.n_def_clusters_updated}",
             "Initialize|Def update time (s)" : f"{self.update_def_time:.6f}",
-            "Predict|Messages cnt" : f"{self.msg_idx+1}",
-            "Predict|Avg predict time (s/msg)" : 
-                f"{self.t_predict/self.n_predict:.6f}" if self.n_predict else "NA",
+            "Predict|Messages cnt" : f"{n_steps}",
+            "Predict|Avg predict time (s/msg)" : f"{self.t_predict/n_steps:.6f}",
             "Predict|Max predict time (s/msg)" : f"{self.max_t_predict:.6f}",
```
jasonrute, line 229: Just so I know, will this be 1 if it is the first message now?

jasonrute (approved):
LGTM!  (I had one question, but it was just so I know the answer when I look at the summary.)

graph2tac/loader/predict_server.py
mirefek: I don't think something would be `1`, what in particular do you mean by "this"?

jasonrute: Oh, sorry, I meant the line below:
```
"Predict|Max time predict msg ix" : f"{self.argmax_t_predict}",
```
If the first predict takes the longest, will it show up as 0 or 1 now?

jasonrute: It being `self.argmax_t_predict`

PR #141, opened 2023-05-04
Set thread count in predict server

jasonrute:
Set the number of threads that tensorflow can use.  I tested that this works by using a job with 2 CPUs but setting the thread count to 1 and using the profiler (which is still in the hack-the-code stage), and saw that the threads being used by TF go down to 1.

### DIFF
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -789,7 +789,7 @@ def load_model(config: argparse.Namespace, log_levels: dict) -> Predict:
         tf.get_logger().setLevel(int(log_levels[config.tf_log_level]))
         tf.config.run_functions_eagerly(config.tf_eager)
         tf.config.threading.set_inter_op_parallelism_threads(
-            config.cpu_thread
+            config.cpu_thread_count
         )
 
         if config.exclude_tactics is not None:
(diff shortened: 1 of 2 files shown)


jasonrute:
The tests are failing.  I need to fix that.  (It is because I can't change the parallelism after TF is installed, but the tests install tensorflow earlier than the rest of the code.  This happens too with the random seed, and I'll fix it the same way.)

graph2tac/loader/predict_server.py
```diff
@@ -760,6 +760,11 @@ def parse_args() -> argparse.Namespace:
                         action='store_true',
                         help="Makes data_server check its inner consistency on each update")
 
+    parser.add_argument('--cpu_thread_count', '--cpu-thread-count',
```
LasseBlaauwbroek, line 763: Why do we have some (but not all) parameters so that we can choose between underscores and dashes? Isn't it easier to just make all parameters use dashes only? Or is there some technical reason?

jasonrute: Historically, Fidel used underscores.  I kept mistyping them, so I added alternate dashed versions.  Also, it can sometimes be confusing since in the python code we have to use the underscore version.  I thought having both is the best compromise.

jasonrute: I'll just make all of them both.  (I don't want to make them all dashes, since I would have to change my scripts.)

jasonrute:
Standardized the arguments, and fixed the tests.  It should be good to go.

LasseBlaauwbroek (approved):

PR #143, opened 2023-05-09
Stop returning values in entry points and improve predict server tests

jasonrute:
Change 1:
No longer return values in `g2t-tfgnn-train`, `g2t-server`, and `g2t-hmodel-train`.  These would be treated as errors and cluttered output.  But the values can still be used for tests.

Change 2:
Change the format of the predict server expected outputs.  There were two problems:
* In our tests, there are many outputs with the same confidence (this is because of symmetry in local arguments).  If we change the algorithm to select the best tactics (like using beam search), then the order could change.  This standardizes it for the test.
* By making the predict server expected outputs a list, it works a bit better when the test fails.  So now, it gives a (more) readable error, like:
```
E         At index 8 diff: {'_type': 'TacticPredictionsGraph', 'contents': {'predictions': []}} != {'_type': 'TacticPredictionsGraph', 'contents': {'predictions': [{'arguments': ['node-1-0-propchain_dep.start'], 'confidence': 0.5 ± 5.0e-05, 'ident': 5294290463909144854}, {'arguments': ['node-1-2-propchain_dep.th0'], 'confidence': 0.5 ± 5.0e-05, 'ident': 5294290463909144854}]}}
```
(Showing in this case that the model returned an empty prediction list for one of the predictions, but we were expecting two predictions.)

### DIFF
--- a/graph2tac/loader/hmodel.py
+++ b/graph2tac/loader/hmodel.py
@@ -230,7 +230,7 @@ class HPredict(Predict):
 
         return np.array(result_pred), result_val
 
-def main():
+def main_with_return_value():
     parser = argparse.ArgumentParser()
 
     parser.add_argument('data_dir', type=Path,
@@ -269,6 +269,10 @@ def main():
     )
 
     return trainer.train()
+
+
+def main():
+    main_with_return_value()
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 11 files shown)


LasseBlaauwbroek (approved):
Looks good! Thanks Jason.

mirefek:
Is the `hmodel-predict-server` supposed to work now? It still doesn't for me but you changed some of the relevant files (so I am not sure what the change is about).

jasonrute:
@mirefek Sorry, I should have made clear, I manually changed all the `expected-predict-server.yaml` files with a script.  That way I could be sure the results are the same except for the permutation of the order the results are listed in (if there is a tie in the confidence).  That is why the hmodel results changed even though the test is still broken.

mirefek:
I see

mirefek (approved):

PR #144, opened 2023-05-09
Add option to profile server

jasonrute:
I added an option to the server to profile a few steps of the prediction and/or the definition task.  Just supply a directory to log to (and optionally the start and stop steps).  I would say it is more of an experimental feature, but it is better to add it to the code than hack it into the code when I need it.

Edit: Usage:
```
g2t-server ... \
--pred-profiler-logdir my_logdir --pred-profiler-start 10 --pred-profiler-end 15 \
--def-profiler-logdir my_logdir2 --def-profiler-start 10 --def-profiler-end 15
```
and then run `tensorboard` on that directory
```
tensorboard --logdir my_logdir
```

### DIFF
--- a/graph2tac/loader/predict_server.py
+++ b/graph2tac/loader/predict_server.py
@@ -17,6 +17,8 @@ import contextlib
 import yaml
 os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
 
+import tensorflow as tf
+
 from pytact.data_reader import (capnp_message_generator, capnp_message_generator_from_file,
                                 TacticPredictionGraph, TacticPredictionsGraph,
                                 GlobalContextMessage, CheckAlignmentMessage, CheckAlignmentResponse,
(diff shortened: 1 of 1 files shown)


graph2tac/loader/predict_server.py
```diff
@@ -17,6 +17,8 @@
 import yaml
 os.environ['TF_CPP_MIN_LOG_LEVEL'] = '3'
 
+import tensorflow as tf
```
jasonrute, line 20: I just added tensorflow at the top as an import.  It is already being imported by `graph2tac.loader.data_classes`, and now it is needed for the profiler.

jasonrute:
The tests all pass (except of the hmodel one which we know fails).

mirefek (approved):
Why not, if you find it useful.

LasseBlaauwbroek:
I do wonder if the profiling itself will slow things down, but lets worry about that later.

jasonrute:
By default it is off.  If you don't supply a directory, it will do basically nothing.  (I mean it will enter a function it immediately exists, but that is nothing compared to the time it takes to run one step of the model.)

jasonrute:
I don't intend the profiling for benchmarking, but for things like I'm doing now, like speeding up the inference model.

PR #146, opened 2023-05-12
Speedup inference model

jasonrute:
This PR *significantly* speeds up the neural network part of inference.  (It still doesn't address beam search which is coming in another PR.)  The idea is a continuation of the previous PR as follows.  After computing the graph embeddings, we choose `k` tactics, and compute arguments for those tactics.  However, the available local and global context is independent from the chosen tactics.  So this allows for three optimizations:
* The keys (local and global context embeddings) can be computed before we do anything with tactics.   They only need to have shape `[batch, None(cxt), hdim]` where `batch` is the original batch size (1 in our case).
* The queries on the other hand, do depend on the tactics, but can still match them up with the keys, by listing all tactic-argument embeddings or all `k` tactics under the same batch element.  The queries then have ragged shape `[batch, None(tactic-args), hdim]`, where `batch` is just the original batch_size (in our case 1).
* Finally, since the batch dimension is size one in our case, the arrays I mentioned above are not really ragged, so in that special case of batch_size=1, we can optimize the multiplication to just be regular matrix multiplication (with no ragged tensor issues).

These three optimizations, when combined, significantly speed up the model.  Previously, most of the time was spent multiplying queries and keys to get the logits (or transforming tensors to get into and out of the right shape to multiply).  This eliminates all that.  And moreover, surprisingly to me, by reducing the shape of the key tensor, we also significantly speed up even the time spent on the einsum operator (even though we still multiply the same number of query-key pairs together).

### DIFF
--- a/graph2tac/tfgnn/tasks.py
+++ b/graph2tac/tfgnn/tasks.py
@@ -971,9 +971,42 @@ class LocalArgumentPrediction(TacticPrediction):
         return tf.gather(hidden_graph.node_sets['node']['hidden_state'],
                                                local_context_ids).with_row_splits_dtype(tf.int64)
 
-    def _hidden_state_sequences(self, hidden_graph: tfgnn.GraphTensor, tactic: tf.Tensor) -> tf.RaggedTensor:
+    def _hidden_state_sequences(
+            self,
+            hidden_graph: tfgnn.GraphTensor,  # [batch]
+            tactic: tf.Tensor,  # [batch]
+        ) -> tf.RaggedTensor:  # [batch, None(args), hdim]
+        # [batch]
         num_arguments = tf.gather(tf.constant(self._graph_constants.tactic_index_to_numargs, dtype=tf.int64), tactic)
-        return self.arguments_head((hidden_graph, self.tactic_embedding(tactic), num_arguments))
+        # [batch, hdim]
+        hidden_state = hidden_graph.context["hidden_state"]
+        # [batch, tactic_hdim]
+        tactic_embedding = self.tactic_embedding(tactic)
+        # [batch, None(args), hdim]
+        return self.arguments_head((hidden_state, tactic_embedding, num_arguments))
+
+    def _hidden_state_sequences_inference(
+        self, 
+        hidden_graph: tfgnn.GraphTensor,  # [batch] 
+        tactic: tf.Tensor,  # [batch, tactic_expand_bound]
+    ) -> tf.RaggedTensor:  # [batch*tactic_expand_bound, None(args), hdim]
+        batch_size = tf.shape(tactic)[0]
+        tactic_expand_bound = tf.shape(tactic)[1]
+
+        tactic = tf.reshape(tactic, shape=[batch_size*tactic_expand_bound])  # [batch*tactic_expand_bound]
+        
+        # [batch*tactic_expand_bound]
+        num_arguments = tf.gather(tf.constant(self._graph_constants.tactic_index_to_numargs, dtype=tf.int64), tactic)
+        
+        hidden_state = hidden_graph.context["hidden_state"]  # [batch, hdim]
+        hidden_state = tf.expand_dims(hidden_state, axis=1)  # [batch, 1, hdim] 
+        hidden_state = tf.tile(hidden_state, multiples=[1, tactic_expand_bound, 1])  # [batch, tactic_expand_bound, hdim]
+        hidden_state = tf.reshape(hidden_state, shape=[batch_size*tactic_expand_bound, tf.shape(hidden_state)[2]])  # [batch*tactic_expand_bound, hdim]
+
+        # [batch*tactic_expand_bound, tactic_hdim]
+        tactic_embedding = self.tactic_embedding(tactic)
+        # [batch*tactic_expand_bound, None(args), hdim]
+        return self.arguments_head((hidden_state, tactic_embedding, num_arguments))
 
     @staticmethod
     def _reshape_inference_logits(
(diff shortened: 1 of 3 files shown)


graph2tac/tfgnn/tasks.py
```diff
@@ -1206,6 +1231,37 @@ def _log_softmax_logits(
...
+        x: tf.RaggedTensor,  # [outer*inner, None(ragged), ...]
+        outer_dim: int,
+        inner_dim: int,
+    ) -> tf.RaggedTensor:  # [inner*outer, None(ragged), ...]
+        """
+        Convert a ragged tensor of shape `[outer*inner, None(ragged), ...]`
+        to one of shape `[inner*outer, None(ragged), ...]`
+        by permuting the inner values.
+        """
+        inner_dim = tf.cast(inner_dim, dtype=tf.int64)
```
jasonrute, line 1245: Tensorflow doesn't support ragged tensors of shape `[a, b, None(c)]`, i.e. where the first two or more dimensions are not ragged.  So I uses few tricks where I combine `a` and `b` into one dimension.  This is a transpose operator for this case.

graph2tac/tfgnn/tasks.py
```diff
@@ -1281,66 +1337,94 @@ def create_inference_model(self, tactic_expand_bound: int, graph_constants: Grap
...
         normalized_global_arguments_logits = self._reshape_inference_logits(logits=normalized_global_arguments_logits,
                                                                            tactic_expand_bound=tactic_expand_bound)
 
         return tf.keras.Model(inputs={self.PROOFSTATE_GRAPH: proofstate_graph, self.TACTIC_MASK: tactic_mask},
                               outputs={self.TACTIC: tf.transpose(top_k_indices),
                                        self.TACTIC_LOGITS: tf.transpose(top_k_values),
-                                       self.LOCAL_ARGUMENTS_LOGITS: normalized_local_arguments_logits,
-                                       self.GLOBAL_ARGUMENTS_LOGITS: normalized_global_arguments_logits})
+                                       self.LOCAL_ARGUMENTS_LOGITS: tf.transpose(normalized_local_arguments_logits, perm=[1, 0, 2, 3]),
+                                       self.GLOBAL_ARGUMENTS_LOGITS: tf.transpose(normalized_global_arguments_logits, perm=[1, 0, 2, 3])})
```
jasonrute, line 1427: It is more natural to work with `[batch, tactic, ...]` instead of `[tactic, batch, ...]`, but the latter is how it was in the code so far.  When I finish beam search, I'll switch it to `[batch, tactic,...]` and get rid of these transpose operations.

graph2tac/tfgnn/tasks.py
```diff
@@ -1281,66 +1337,94 @@ def create_inference_model(self, tactic_expand_bound: int, graph_constants: Grap
...
-
-        hidden_state_sequences = self._hidden_state_sequences(hidden_graph=hidden_graph, tactic=tactic)
         # [tactic_expand_bound*batch, None(args), hdim]
+        hidden_state_sequences = self._hidden_state_sequences(hidden_graph=hidden_graph, tactic=tactic)
+        # [batch*tactic_expand_bound, None(args), hdim]
+        hidden_state_sequences = self.transpose_ragged_first_dim(
+            hidden_state_sequences, 
+            outer_dim=tactic_expand_bound, 
+            inner_dim=batch_size
+        )
```
jasonrute, line 1370: This is a bit awkward.  We stack all the graphs for the array in order to compute the tactic hidden embeddings, but really, there isn't a need to do this stacking, especially since it puts the tactic dimension first.  I probably can rewrite `_hidden_state_sequences`, but it wasn't a major bottleneck, so I didn't get to it.

graph2tac/tfgnn/tasks.py
```diff
@@ -1206,6 +1231,37 @@ def _log_softmax_logits(
...
+
+        # calculate new row ids
+        rowids = x.value_rowids()  # [outer-inner-ragged]
+        batch1_rowids = rowids // inner_dim  # [outer-inner-ragged]
+        batch2_rowids = rowids % inner_dim   # [outer-inner-ragged]
+        new_rowids = batch2_rowids * outer_dim + batch1_rowids  # [outer-inner-ragged]
+        
+        # reorder the elements according to the new row ids
+        new_order = tf.argsort(new_rowids)  # [inner-outer-ragged]
+        new_rowids = tf.sort(new_rowids)  # [inner-outer-ragged]
```
mirefek, line 1256: Couldn't the following lines be written just with some `tf.gather` / `tf.invert_permutation`? (not that I would understand fully the nuances of the code)

jasonrute: I am not sure tf.invert_permute works here.  new_rowids isn't a permutation.  (Indeed there can be multiple batch elements with the same row ids.  I'm specifically using that tf.argsort maintains order when there is a tie.)  I guess instead of tf.sort I can could use gather on the sorted indices.  I don't know what is faster.  But also, this isn't really a slow operation in the code (and maybe I should just fix the need for having it in the first place).

jasonrute: Oh, I need to set stable=true for that in the argsort.

jasonrute:
Actually, I decided to just do things right and fix up `_hidden_state_sequences` to use the original graph instead of the stacked graph.  Then we can avoid stacking graphs just to get out one tensor.  We can always have the batch dimension first, avoiding the need for the ragged transpose function.  Can you look this over again @mirefek now that the code is a bit different?

graph2tac/tfgnn/tasks.py
jasonrute: I decided to address this directly.

mirefek (approved):
Makes sense. I am not checking all the details but the overall idea looks good, and I am happy to merge.

PR #147, opened 2023-05-12
Mirek/graph constants json

mirefek:
As discussed, saves / loads `graph_constants` as `json`. Comments:
* There is a potential issue with the test `load_previous_model` -- since the current code saves `json` in case of loading `yaml`, it will save this `json` to this test, so the next run of the test is not testing as far reaching backward compatibility as before. If we want to be more rigorous about this, we could delete `graph_constants.json` after / before each run of the test.
* I moved discarding possible redundant `global_context` from `data_classes.py / GraphConstants` to the point where it is loaded from `yaml` / converted to `json`.

### DIFF
--- a/graph2tac/tfgnn/predict.py
+++ b/graph2tac/tfgnn/predict.py
@@ -163,7 +163,7 @@ class TFGNNPredict(Predict):
         if not graph_constants_filepath_json.exists():
             logger.info(f'no json graph_constants file, trying to load yaml')
             with graph_constants_filepath_yaml.open('r') as yml_file:
-                graph_constants_d = yaml.load(yml_file, Loader=yaml.UnsafeLoader)
+                graph_constants_d = yaml.load(yml_file, Loader=yaml.SafeLoader)
             if "global_context" in graph_constants_d:
                 global_context = graph_constants_d["global_context"]
                 assert global_context == list(range(len(global_context)))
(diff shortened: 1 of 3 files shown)


graph2tac/tfgnn/predict.py
```diff
@@ -157,11 +158,26 @@ def __init__(self,
...
-        with graph_constants_filepath.open('r') as yml_file:
-            graph_constants_d = yaml.load(yml_file, Loader=yaml.UnsafeLoader)
-            graph_constants_d['data_config'] = DataConfig(**graph_constants_d['data_config'])
-            graph_constants = GraphConstants(**graph_constants_d)
+        graph_constants_filepath_yaml = log_dir / 'config' / 'graph_constants.yaml'
+        graph_constants_filepath_json = log_dir / 'config' / 'graph_constants.json'
+        if not graph_constants_filepath_json.exists():
+            logger.info(f'no json graph_constants file, trying to load yaml')
+            with graph_constants_filepath_yaml.open('r') as yml_file:
+                graph_constants_d = yaml.load(yml_file, Loader=yaml.UnsafeLoader)
```
jasonrute, line 166: Can you make this SafeLoader?  All our yaml should be safe now and this prevents code execution.

jasonrute (approved):
Looks good to me.  If the creation of the json file during tests is annoying, we could later do something about it, but overall it is useful.  Just one small request about loading the yaml.  If you fix that I'm happy with you merging it.

graph2tac/tfgnn/predict.py
mirefek: ok, I just copied the original yaml-loading code, I will change it to SafeLoader

PR #148, opened 2023-05-22
Speed up inference considerably

jasonrute:
This big PR speeds up the inference considerably.  It contains a lot of new things, so I'm going to try to document it clearly.  I still have some code cleanup to do, but I think it is better to make the PR so others can start to look at it.

Here are the big parts:
* *A layer `BeamSearch`.*  I couldn't find one in tensorflow or related packages which was fast enough and fit my needs, so I implemented my own.  It is simple, but very fast for our purposes.  It is also general enough that we could easily do more tranditional beam search where say we first choose the tactic, and then choose the first argument, and then choose the next argument, all using the previous argument.
* *A layer `SelectBestResults`* which selects the best tactics using the beam search layer.  It also transforms the tensors into a form needed for beam search as well as undos those transformations to put the results into the right format (with the right indices).  I've tried to comment this really well, but at the same time it is a lot of indexing into indexes.  I think I want to slightly reformat the code to make it even easier to follow by making it a nested set of functions where each function transformers the arrays, calls another function, and then transforms the arrays back.  That way the encoding and decoding are right next to each other.
* *I moved a lot of stuff out of the base inference model*, including softmax, and turning the arrays into dense arrays.  Softmax is faster to implement on dense tensors when I've already concatenated the local and global arguments.  I make the context dimension dense (which makes sense especially if it is a batch size of one, since then it already is basically dense since every tactic gets it's own batch).
* *I made the inference code flexible enough to use the full batch size* (although I didn't go as far as finishing implementing batch_prediction), but I also optimized for a batch size of one.  There are a few tricks with a batch size of one that really help make it faster, especially since right now that is the only way we are using it.

Here are some smaller changes:
* *There is no need for `total_expand_bound`.*  I've left it in the argument list to not break scripts, but it doesn't do anything anymore.
* *The base inference model as well as the beam search are all compiled together.*  This makes the transition between the two instantaneous.  To do this, search_expand_bound is now compiled into inference.  (If we ever need a flexible search expand bound that as passed as an argument to predict, that wouldn't be hard to implement, but this seemed to make sense since it is a constant.)
* *Modified the expected results.*  Unfortunately, our tests contain a number of symmetric local arguments.  The beam search orders the top results, but since we only return 4 top results, we sometimes have other results with the same score that we skipped.  To make sure it isn't just an off by one index thing I verified that in some cases the results doesn't change.  And when they do, the probabilities are the same.  To generate the new expected results files, I wrote a script which only changed the arguments in the files without changing the probabilities.  So we can see the better what changed.

Stuff I'd like to do before merging:
- [ ] Remove the code I forgot to remove.
- [ ] Document stuff a bit better
- [ ] See if removing the early stopping from the beam search matters.  It might even make it faster since the early stopping doesn't do much, and it would simplify the code a lot.
- [ ] Better structure `SelectBestResults`
- [ ] Implement batch_prediction if really easy.  (But probably not.)

### DIFF
--- a/graph2tac/predict.py
+++ b/graph2tac/predict.py
@@ -64,17 +64,20 @@ class Predict:
     def __init__(self,
                  graph_constants: GraphConstants,
                  tactic_expand_bound: int,
+                 search_expand_bound: int,
                  debug_dir: Optional[Path] = None,
     ):
         """
 
         @param graph_constants: the graph constants seen during training
-        @param total_expand_bound:
+        @param total_expand_bound: how many base tactics to select (may be ignored by the subclass)
+        @param search_expand_bound: how many total tactics to return
         @param debug_dir: a directory where all api calls will be logged for debugging purposes
         """
         self.graph_constants = graph_constants 
         self._debug_dir = debug_dir
         self._tactic_expand_bound = tactic_expand_bound
+        self._search_expand_bound = search_expand_bound
         self._timings = {}
         assert self.graph_constants is not None
 
@@ -105,7 +108,6 @@ class Predict:
                            state: LoaderProofstate,
                            allowed_model_tactics: List[int],
                            available_global: Optional[np.ndarray],
-                           total_expand_bound: int
                            ) -> Tuple[np.ndarray, List]:
         """
         [ Public API ] Returns actions ordered by their corresponding probabilities, as computed by the model.
@@ -116,7 +118,6 @@ class Predict:
         @param state: (graph, root, context)
         @param allowed_model_tactics:
         @param available_global: np.array of indices into global_context
-        @param total_expand_bound:
         @return: a pair (ranked_actions, ranked_values)
         """
         raise NotImplementedError('ranked_predictions should be implemented by sub-classes')
--- a/graph2tac/tfgnn/predict.py
+++ b/graph2tac/tfgnn/predict.py
@@ -21,14 +21,20 @@ from graph2tac.tfgnn.stack_graph_tensors import stack_graph_tensors
 
 
 class BeamSearch(tf.keras.layers.Layer):
-    def __init__(self, logits_fn, stopping_fn, eos_id: int, max_decode_length: int):
+    def __init__(self, logits_fn, stopping_fn, eos_id: int):
         super().__init__()
         self.logits_fn = logits_fn
         self.stopping_fn = stopping_fn
         self.eos_id = eos_id
         #self.max_beam_size = max_beam_size,
         # this needs to be a python int to make sure that the beam search knows how many steps of the loop to compile
-        self.max_decode_length = int(max_decode_length)  
+    
+    @staticmethod
+    def expand_dims_none(x: tf.Tensor, axis: int):
+        "Like tf.expand_dims, but sets the new axis dimension to be None instead of 1"
+        x = tf.expand_dims(x, axis=axis)
+        # set axis to None with this trick
+        return tf.reshape(x, shape=tf.tensor_scatter_nd_update(tf.shape(x), [[axis % tf.rank(x)]], [-1]))
     
     def one_beam_step(
         self,
@@ -73,9 +79,10 @@ class BeamSearch(tf.keras.layers.Layer):
         }
 
         return ids, top_k_scores, cache
-
+    
     def one_beam_step_with_early_stopping(
         self,
+        is_finished: tf.Tensor, # bool
         ids: tf.Tensor,  # [batch_size, beam_size0, seq_length]
         scores: tf.Tensor,  # [batch_size, beam_size0]
         cache: dict[str, tf.Tensor],  # dict with values: [batch_size, beam_size0, ...]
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -23,6 +23,8 @@
         'keras>=2.8',
         'tensorflow>=2.9.0,<2.10', # TODO(jrute): Try to upgrade to >=2.10
         'tensorflow_gnn>=0.2.0,<0.3',  # TODO(jrute): Try to upgrade to >=0.3
+        'tf-models-official',
+        'keras-nlp',
```
jasonrute, line 27: Note to self.  Remove this.  (Left over from trying premade beam search implementations.)

tests/data/duplicate_identities/params/predict_server/predict_server.yml
```diff
@@ -1,7 +1,7 @@
 arch: tfgnn
 log_level: info
 tf_log_level: critical
-tactic_expand_bound: 2  # there are only 3 tactics in this dataset
+tactic_expand_bound: 5  # there are only 3 tactics in this dataset
```
jasonrute, line 4: I changed this to 5 to test that we can now make the tactic_expand_bound bigger than that number of available tactics.   This sort of thing used to crash.  It doesn't change the results.

graph2tac/tfgnn/tasks.py
```diff
@@ -1356,66 +1361,50 @@ def create_inference_model(self, tactic_expand_bound: int, graph_constants: Grap
...
         global_arguments_logits = self.global_logits(queries=global_hidden_state_sequences, keys=global_embeddings)
-        # [batch*tactic_expand_bound, None(args), None(context)]
+        # [batch*top_k_tactics, None(args), None(context)]
         global_arguments_logits = tf.RaggedTensor.from_row_lengths(
             values=global_arguments_logits.values,  # [batch-tactic-arg, None(context)]
             row_lengths=tactic_arg_cnt
         )
-        
-        # [batch*tactic_expand_bound, None(args), None(context)], [batch*tactic_expand_bound, None(args), None(context)]
-        normalized_local_arguments_logits, normalized_global_arguments_logits = self._log_softmax_logits(
```
jasonrute, line 1398: Softmax is done later when the tensors are in a better shape

graph2tac/tfgnn/predict.py
```diff
@@ -19,6 +19,384 @@
...
+        self.eos_id = eos_id
+        #self.max_beam_size = max_beam_size,
+        # this needs to be a python int to make sure that the beam search knows how many steps of the loop to compile
+    
+    @staticmethod
+    def expand_dims_none(x: tf.Tensor, axis: int):
+        "Like tf.expand_dims, but sets the new axis dimension to be None instead of 1"
+        x = tf.expand_dims(x, axis=axis)
+        # set axis to None with this trick
+        return tf.reshape(x, shape=tf.tensor_scatter_nd_update(tf.shape(x), [[axis % tf.rank(x)]], [-1]))
```
jasonrute, line 37: Note to self.  I probably don't need this anymore.  It was a bit of a hack.

graph2tac/tfgnn/predict.py
```diff
@@ -19,6 +19,384 @@
 from graph2tac.tfgnn.graph_schema import vectorized_definition_graph_spec, proofstate_graph_spec, batch_graph_spec
 from graph2tac.tfgnn.stack_graph_tensors import stack_graph_tensors
 
+
+class BeamSearch(tf.keras.layers.Layer):
```
jasonrute, line 23: TODO: Document how the BeamSearch class works.

graph2tac/tfgnn/predict.py
```diff
@@ -423,7 +811,9 @@ def ranked_predictions(self,
         if available_global is not None:
             raise NotImplementedError('available_global is not supported yet')
 
-        inference_output = self._inference_model(state, allowed_model_tactics)
+        best_log_probs, best_tactics = self._inference_model(state, allowed_model_tactics)
+        return (best_tactics.numpy()[0], np.exp(best_log_probs.numpy()[0]))
+
         predict_output = PredictOutput(state=None, predictions=[])
 
         # go over the tactic_expand_bound batches
```
jasonrute, line 819: TODO: Remove all the code below this.  (And probably lots of other code. :smile:)

setup.py
jasonrute: Fixed.

graph2tac/predict.py
```diff
@@ -64,17 +64,20 @@ class Predict:
...
                  graph_constants: GraphConstants,
                  tactic_expand_bound: int,
+                 search_expand_bound: int,
                  debug_dir: Optional[Path] = None,
     ):
         """
 
         @param graph_constants: the graph constants seen during training
-        @param total_expand_bound:
+        @param total_expand_bound: how many base tactics to select (may be ignored by the subclass)
```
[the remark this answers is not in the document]
jasonrute: Fixed.

jasonrute:
I addressed most of the comments above and I also took the opportunity to delete most of the dead code. 😎

PR #150, opened 2023-05-31
Use log prob

jasonrute:
Switch to using returning log probability instead of probability.

I manually converted the tests to log to make sure the numbers are still correct.


jasonrute:
I would normally wait for a review, but I'm just going to merge this so that we have main how we would like it for benchmarking.  I think this is well tested.

PR #154, opened 2023-11-06
Upgrade to TF 2.12 and support python 3.11

jasonrute:
This PR does the following:
* Upgrades to TF 2.12
* Adds support for Python 3.11
* Drops support for Python 3.8 (since not supported by TF 2.10+)

The test results had to change going from Tensorflow 2.9 to 2.10.  This is because Tensorflow 2.10 changed the way Tensorflow initialized the weights.  I think it is the same default initialization distribution, but the implementation is different.  In particular setting the initialization also uses the python and/or numpy random number generators.  So it is like running the tests with different seed.

Unfortunately, we don't have regression tests so it isn't clear that we didn't introduce a bug in the upgrade which made things worse.  I doubt it, but note that in our very small and short test runs the accuracies changed widely, sometimes up and sometimes down.

The good news is that the tests are still deterministic (working on both mac with venv and linux with conda), the same expected results worked when upgrading from Tensorflow 2.10 to 2.12.  And importantly, a model trained with a former version of this repository still gives the same results.


jasonrute:
I have noticed that the val_tactic_logits_accuracy never goes up, but sometimes goes down a lot.  The `val_tactic_logits_loss` doesn't have as drastic of a change.  It could likely be random noise, but maybe I'll run a longer test run quick on ministdlib to be sure it is fine.

jasonrute:
I did a longer run with TF 2.12 (and Python 3.11).  It works fine and the results are basically the same (up to random noise).  So merging this...

LasseBlaauwbroek:
Thanks @jasonrute !

